PR #95, opened 2015-06-29
Uploaded length of multipart put should match user specified length. Closes #94

### LINKED ISSUE #94: handle cases where incoming put data is too large or too small

fkautz:


fkautz:
LGTM

PR #102, opened 2015-06-29
URL builder now escapes queries. Closes #84

### LINKED ISSUE #84: Check all query parameters are properly escaped when building url
Add code to escaspe, including / in query params

fkautz:


fkautz:
LGTM

PR #114, opened 2015-07-01
Returned timestamps for bucket and object are now datetime objects instead of string. Closes #107.

### LINKED ISSUE #107: Parse dates and return datetime objects from parsed xml

fkautz:


fkautz:
LGTM

PR #156, opened 2015-07-08
list objects uses last key is used if marker is not set. Fixes #152

### LINKED ISSUE #152: Use last object if NextMarker is not set and IsTruncated is true
NextMarker may not be populated, use the last object seen if this is the case.

fkautz:


fkautz:
LGTM

PR #174, opened 2015-07-09
get_bucket_acl now checks for errors properly. Fixes #162.

### LINKED ISSUE #162: get_bucket_acl of a non existent bucket returns 'private'
Should throw an exception stating bucket doesn't exist instead.

fkautz:


fkautz:
LGTM

PR #182, opened 2015-07-09
Return an error on data size mismatch. Fixes #171.

### LINKED ISSUE #171: Putting a small object that is too large or too small should fail before sending request
Both of these should fail before sending request:
```py
# put an object with length set too small.
client.put_object(bucket, 'string_object', 10, 'hello world')

# put an object with length set too large
client.put_object(bucket, 'string_object', 12, 'hello world')
```

fkautz:


fkautz:
LGTM

PR #192, opened 2015-07-09
Return a more descriptive error type on size mismatch. Fixes #189.

### LINKED ISSUE #189: multipart put returns ValueError rather than DataSizeMismatch
11m_text is an 11mb text file.

```py
with open('/home/fkautz/11m_text', 'r') as file:
    client.put_object(bucket, '11mb_text_stream', 11*1024*1024+1, file)
```

fkautz:


fkautz:
LGTM

PR #201, opened 2015-07-10
Fix delimiter being double encoded. Fixes #199.

### LINKED ISSUE #199: Recursion sends delimiter=%252F instead of %2F
Double urlencoding is occurring.

```
%25 => %
%2F => /
```

fkautz:


fkautz:
LGTM

PR #253, opened 2015-08-04
Fixes needed because of the move from request to urllib3

krishnasrinivas:


krishnasrinivas:
let me check.

krishnasrinivas:
Looks like a bug on requests (https://github.com/kennethreitz/requests/issues/1561) and urllib3 (https://github.com/shazow/urllib3/issues/236)

harshavardhana:
Interesting, did not know when they changed this behavior.

harshavardhana:
Perhaps its requests v/s urllib3

krishnasrinivas:
@harshavardhana requests also has the same behavior. Maybe we used different version of request during development? check here github.com/kennethreitz/requests/issues/1561

harshavardhana:
> @harshavardhana requests also has the same behavior. Maybe we used different version of request during development? check here github.com/kennethreitz/requests/issues/1561

yeah strange indeed, i wonder if there were working properly at all :(

PR #304, opened 2015-12-04
signature: Add more signature v4 documentation and cleanup.

harshavardhana:


minio/error.py
```diff
@@ -18,6 +18,10 @@
 ~~~~~~~~~~~~~~~~~~~
 
 This module provides custom exception classes for Minio library and API specific errors.
```
balamurugana: I think you need to run pep8

minio/helpers.py
```diff
@@ -123,21 +166,22 @@ def is_valid_endpoint(endpoint):
...
+
+    allowed_hostname = re.compile(
         "^((?!-)[A-Z\\d-]{1,63}(?<!-)\\.)*((?!-)[A-Z\\d-]{1,63}(?<!-))$",
         re.IGNORECASE)
-    if not allowed.match(hostname):
-        raise InvalidEndpointError('endpoint')
+    if not allowed_hostname.match(hostname):
+        raise InvalidEndpointError('Hostname does not meet URL standards.')
 
     if hostname.endswith('amazonaws.com') and (hostname != 's3.amazonaws.com'):
```
balamurugana: it would be better to have ".amazonaws.com" than "amazonaws.com"

minio/helpers.py
```diff
@@ -123,21 +166,22 @@ def is_valid_endpoint(endpoint):
...
-        raise InvalidEndpointError('endpoint')
+        raise InvalidEndpointError('Hostnames for Amazon S3 should always be s3.amazonaws.com')
```
balamurugana: I feel "Amazon S3 hostname should be s3.amazonaws.com" string is better.  Is it good to have hostname in the exception?

minio/helpers.py
```diff
@@ -156,11 +200,11 @@ def is_valid_bucket_name(bucket_name):
...
     """
     ## verify bucket name length.
-    if len(bucket_name) < 3: 
+    if len(bucket_name) < 3:
         raise InvalidBucketError('Bucket name cannot be less than 3 characters.')
     if len(bucket_name) > 63:
         raise InvalidBucketError('Bucket name cannot be more than 63 characters.')
-    
+
     validbucket = re.compile('^[a-zA-Z][a-zA-Z0-9\\-]+[a-zA-Z0-9]$')
```
balamurugana: Please compile regex at only once.  Here it will be compiled every function call.

balamurugana:
Few minor comments.

minio/helpers.py
harshavardhana: > Is it good to have hostname in the exception?

Should be fine i guess.

minio/helpers.py
harshavardhana: Correct, let me make it global.  There is one more place this is needed.

minio/error.py
harshavardhana: Will run it.

harshavardhana:
> Few minor comments.

Fixing it.

harshavardhana:
pep8 related changes, will bring in at a later stage. Opening a relevant bug against it.

minio/helpers.py
```diff
@@ -109,6 +152,10 @@ def get_target_url(endpoint, bucket_name=None, object_name=None, query=None):
 
     return ''.join(url_components)
 
+allowed_hostname = re.compile(
+    "^((?!-)[A-Z\\d-]{1,63}(?<!-)\\.)*((?!-)[A-Z\\d-]{1,63}(?<!-))$",
+    re.IGNORECASE)
```
balamurugana: You need to move this to global level as well.

harshavardhana: > You need to move this to global level as well.

This is already at the global level.

balamurugana: Sorry, I couldn't able to notice that.  You could move globals into beginning after the imports and rename then like _ALLOWED_HOSTNAME_REGEX.

harshavardhana: Ah let me do that.

minio/helpers.py
```diff
@@ -25,6 +31,10 @@
 from .compat import urlsplit, basestring, urlencode
 from .error import InvalidBucketError, InvalidEndpointError
 
+_VALID_BUCKETNAME_REGEX = re.compile('^[a-zA-Z][a-zA-Z0-9\\-]+[a-zA-Z0-9]$')
+_ALLOWED_HOSTNAME_REGEX = re.compile("^((?!-)[A-Z\\d-]{1,63}(?<!-)\\.)*((?!-)[A-Z\\d-]{1,63}(?<!-))$",
+                                     re.IGNORECASE)
+
```
balamurugana: Thanks.

balamurugana:
+1 LGTM

PR #306, opened 2015-12-04
api: Cleanup another round

harshavardhana:
- split _do_put_object into _do_put_object and _do_put_multipart_object.
- Fix couple of bugs in streaming upload.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -25,7 +25,7 @@ This module implements the API.
 
 """
 
-### Standard python packages
+# Standard python packages
 import sys
 import io
 import platform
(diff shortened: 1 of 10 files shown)


minio/api.py
```diff
@@ -839,29 +874,49 @@ def _do_put_object(self, bucket_name, object_name, data,
...
         uploaded_etags = []
         total_uploaded = 0
         current_part_number = 1
 
-        ## If upload_id is None its a new multipart upload.
+        # Some streams wouldn't implement seek() method.
+        # Handle it and set it to False.
+        try:
+            data.seek(0)
+        except:
```
balamurugana: its not a good idea to ignore any exception.

harshavardhana: > its not a good idea to ignore any exception.

It was on purpose to know if a particular method exists, if it doesn't then we fall back to a different method to upload. I will find a better way to know this.

balamurugana:
+1 LGTM

PR #326, opened 2015-12-17
listIncompleteUploads: Fix argument overloading bug, which causes man…

harshavardhana:
…gling of inputs.

This was done in a previous patch which caused issues in put_object.


balamurugana:
:+1: LGTM

PR #344, opened 2016-01-01
parsers: Improve parsers to use find and findall xpath style

harshavardhana:


balamurugana:
As per our previous discussion on this.  gluster-nagios-common proj has xml2dict
```python
def xml2dict(tree):
    d = {tree.tag: {} if tree.attrib else None}
    children = list(tree)
    if children:
        dd = defaultdict(list)
        for dc in map(xml2dict, children):
            for k, v in dc.iteritems():
                dd[k].append(v)
        d = {tree.tag: {}}
        for k, v in dd.iteritems():
            d[tree.tag][k] = v[0] if len(v) == 1 else v
    if tree.attrib:
        d[tree.tag].update((k, v) for k, v in tree.attrib.iteritems())
    if tree.text:
        text = tree.text.strip()
        if children or tree.attrib:
            if text:
                d[tree.tag]['#text'] = text
        else:
            d[tree.tag] = text
    return d
```

for large xml, this code unnecessarily parsing entire xml which might not be required i feel instead we would need to select the value what we want.
```python
tree.find('volStatus/volumes/volume/volName').text
```

if dict is must have, you could extend dict to do lazy parsing

harshavardhana:
> if dict is must have, you could extend dict to do lazy parsing

Dictionary is not a hard requirement, it is good to have though.

balamurugana:
:+1: LGTM

minio/parsers.py
```diff
@@ -58,34 +59,20 @@ def parse_multipart_upload_result(data):
...
-        raise InvalidXMLError('Missing "ETag" XML attribute.')
-
-    return MultipartUploadResult(bucket_name, object_name, location, etag)
+    try:
+        bucket_name = root.find('s3:Bucket', _S3_NS).text
+        object_name = root.find('s3:Key', _S3_NS).text
+        location = root.find('s3:Location', _S3_NS).text
+        etag = root.find('s3:ETag', _S3_NS).text
+        # Strip off quotes from begining and the end.
+        if etag.startswith('"') and etag.endswith('"'):
```
balamurugana: you could have
```python
etag = etag.replace('"', '')
```

minio/parsers.py
```diff
@@ -58,34 +59,20 @@ def parse_multipart_upload_result(data):
...
-    if not object_name:
-        raise InvalidXMLError('Missing "Key" XML attribute.')
-    if not location:
-        raise InvalidXMLError('Missing "Location" XML attribute.')
-    if not etag:
-        raise InvalidXMLError('Missing "ETag" XML attribute.')
-
-    return MultipartUploadResult(bucket_name, object_name, location, etag)
+    try:
+        bucket_name = root.find('s3:Bucket', _S3_NS).text
```
balamurugana: is it good idea to have a func getText(root, xpath, ns=_S3_NS) ?

balamurugana:
:+1: LGTM

PR #345, opened 2016-01-04
trace: Redact signature header value from being printed in trace logs.

harshavardhana:
Also print response body during error xml.


balamurugana:
:+1: LGTM

PR #355, opened 2016-01-12
get_bucket_location: Do not default to 'us-east-1' for non S3 endpoints.

harshavardhana:
We do not need to default to 'us-east-1' for non-S3 endpoints.


balamurugana:
:+1: LGTM

PR #357, opened 2016-01-14
put_object: Now default to using total_parts_count.

harshavardhana:


minio/api.py
```diff
@@ -455,23 +452,25 @@ def fput_object(self, bucket_name, object_name, file_path,
 
         # Always start with first part number.
         part_number = 1
-        while True:
-            # If EOF reached, break out of the loop.
-            if total_uploaded == file_size:
-                break
+        while part_number <= total_parts_count:
```
balamurugana: As you need to manage to increment part_number by 1, i would prefer to have
```python
for part_number in xrange(1, total_part_count):
```

minio/api.py
```diff
@@ -1302,36 +1314,37 @@ def _stream_put_object(self, bucket_name, object_name,
...
-            if total_uploaded == data_content_size:
-                break
-
-            part_metadata = parts_manager(data)
-            data_md5_hex = encode_to_hex(part_metadata.md5digest)
+        # Generate new parts and upload <= current_part_size until
+        # part_number reaches total_parts_count calculated for the
+        # given size. Additionally part_manager() also provides
+        # md5digest and sha256digest for the partitioned data.
+        while part_number <= total_parts_count:
```
balamurugana: same as above

minio/api.py
harshavardhana: Ah this is python.

balamurugana:
:+1: LGTM

PR #363, opened 2016-02-19
Fix for bucket_exists

dgelvin:


dgelvin:
Should test for equality, not identity.

harshavardhana:
> Should test for equality, not identity.

Thanks

PR #370, opened 2016-05-16
`xrange` is not available in Py3

dhilipsiva:
More Py3 compatibiltiy


harshavardhana:
LGTM - thanks.

dhilipsiva:
@harshavardhana Sorry to bother - but any idea when the new version will be pushed to PyPI server?

PR #371, opened 2016-05-16
Count till Nth than before Nth

dhilipsiva:
`range(1,N)` function returns `[1, 2 ,..., N-1]`. So the line `if part_number == total_parts_count` never evaluates to `True`. It's always `False`. I tried changing it to `if part_number == total_parts_count - 1`, but it did not work. I changed the range to `range(1, N+1)`. It seems to be working for me.  ***BUT***, this might not be an Ideal fix. Just sending a PR to whatever scratched my itch.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -391,7 +391,7 @@ class Minio(object):
             uploaded_parts[part.part_number] = part
 
         # Always start with first part number.
-        for part_number in range(1, total_parts_count):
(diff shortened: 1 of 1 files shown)


dhilipsiva:
Which makes me wonder - how come no one raised this Issue before? 😛

harshavardhana:
This is a correct fix.

PR #383, opened 2016-07-08
doc:API.md/Modifications to bring it closer to docs.minio.io

koolhead17:


harshavardhana:
There are two commits in this patch, should be only one.

PR #400, opened 2016-08-28
api: Fix python3 compatibility issue by using compat values.

### LINKED ISSUE #399: Verify with python 3.5.1
@xtream1101 reported issue with set policy in python 3.5.1:

```
File "/Users/eddyhintze/Virtualenvs/py3/py3/lib/python3.5/site-packages/minio/api.py", line 361, in set_bucket_policy
    bucket_name, prefix)
  File "/Users/eddyhintze/Virtualenvs/py3/py3/lib/python3.5/site-packages/minio/policy.py", line 496, in set_policy
    append_statements(out, ns)
  File "/Users/eddyhintze/Virtualenvs/py3/py3/lib/python3.5/site-packages/minio/policy.py", line 372, in append_statements
    _append_statement(statements, s)
  File "/Users/eddyhintze/Virtualenvs/py3/py3/lib/python3.5/site-packages/minio/policy.py", line 356, in _append_statement
    _get_action(s) == _get_action(statement) and
  File "/Users/eddyhintze/Virtualenvs/py3/py3/lib/python3.5/site-packages/minio/policy.py", line 57, in _get_action
    if isinstance(action, basestring):
NameError: name 'basestring' is not defined
```

```
  File "/Users/eddyhintze/Virtualenvs/py3/py3/lib/python3.5/site-packages/minio-2.0.0-py3.5.egg/minio/api.py", line 372, in set_bucket_policy
  File "/Users/eddyhintze/Virtualenvs/py3/py3/lib/python3.5/site-packages/minio-2.0.0-py3.5.egg/minio/helpers.py", line 417, in get_md5
TypeError: Unicode-objects must be encoded before hashing
```

```
File "/Users/eddyhintze/Virtualenvs/py3/py3/lib/python3.5/site-packages/scrapers/__init__.py", line 124, in <module>
    'data')
  File "/Users/eddyhintze/Virtualenvs/py3/py3/lib/python3.5/site-packages/minio-2.0.0-py3.5.egg/minio/api.py", line 372, in set_bucket_policy
TypeError: string argument without an encoding
```

harshavardhana:
Also fix the API set_bucket_policy() should take input
policy as the last option and API should be consistent
with other languages.

This patch also adds a functional test case.

Fixes #399

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -311,7 +311,7 @@ class Minio(object):
                                       bucket_name=bucket_name,
                                       query={"policy": ""},
                                       headers={})
-            policy_dict = json.loads(response.data)
+            policy_dict = json.loads(response.read().decode('utf-8'))
         except ResponseError as e:
             # Ignore 'NoSuchBucketPolicy' error.
             if e.code != 'NoSuchBucketPolicy':
@@ -336,7 +336,7 @@ class Minio(object):
 
         return policy.get_policy(statements, bucket_name, prefix)
 
-    def set_bucket_policy(self, policy_access, bucket_name, prefix=""):
+    def set_bucket_policy(self, bucket_name, prefix, policy_access):
         """
         Set bucket policy of given bucket name and object prefix.
 
@@ -368,9 +368,11 @@ class Minio(object):
             policy_dict['Statement'] = statements
             content = json.dumps(policy_dict)
 
-            headers = {'Content-Length': str(len(content)),
-                       'Content-MD5': encode_to_base64(get_md5(content))}
-            content_sha256_hex = encode_to_hex(get_sha256(content))
+            headers = {
+                'Content-Length': str(len(content)),
+                'Content-MD5': encode_to_base64(get_md5(content.encode('utf-8')))
+            }
+            content_sha256_hex = encode_to_hex(get_sha256(content.encode('utf-8')))
 
             self._url_open("PUT",
                            bucket_name=bucket_name,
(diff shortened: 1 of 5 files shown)


harshavardhana:
ping @balamurugana

minio/api.py
```diff
@@ -336,7 +336,8 @@ def get_bucket_policy(self, bucket_name, prefix=""):
 
         return policy.get_policy(statements, bucket_name, prefix)
 
-    def set_bucket_policy(self, policy_access, bucket_name, prefix=""):
+    def set_bucket_policy(self, bucket_name, prefix="",
+                          policy_access=policy.Policy.NONE):
```
balamurugana: Its good to not to have `policy_access` is an optional argument.

harshavardhana: Optional is because its a default case is remove bucket policy.

balamurugana: Why do we need remove as default? In most of the cases a valid bucket policy is set.  Having `Policy.NONE` as default would lead to loosing the policy when the API is called with wrong code.  I feel making `policy_access` mandate to avoid such errors.

harshavardhana: > Why do we need remove as default? In most of the cases a valid bucket policy is set. Having Policy.NONE as default would lead to loosing the policy when the API is called with wrong code. I feel making policy_access mandate to avoid such errors.

That we can do and not have any optional params for this API.

balamurugana:
LGTM :+1:

PR #406, opened 2016-09-29
api: Implement copyObject API.

harshavardhana:


harshavardhana:
Can you review this @balamurugana ?

balamurugana:
minio/signer.py:239, we would need to fix query params here by sorting only by the key.

harshavardhana:
> minio/signer.py:239, we would need to fix query params here by sorting only by the key.

This is fixed in a separate patch in listen bucket notifications.

harshavardhana:
This patch is ready to go.

balamurugana:
> This patch is ready to go.

Ok.  Merging the patch now.

PR #407, opened 2016-10-02
Implement listen_bucket_notification

harshavardhana:


examples/set_bucket_policy.py
```diff
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
-# Minio Python Library for Amazon S3 Compatible Cloud Storage,
-# (C) 2016 Minio, Inc.
+# Minio Python Library for Amazon S3 Compatible Cloud Storage, (C) 2016 Minio, Inc.
```
balamurugana: Comment needs to be 79 char long as per pep8 :)

harshavardhana: This is a license header @balamurugana we shouldn't change this because pep8 lacks understanding. It is not a correct practice.

balamurugana: Python's FAQ on this says something like below
```
Copyright (c) <year>, <your name>
```

harshavardhana: Done @balamurugana

PR #423, opened 2016-11-22
Fix a bug in presigned URL, resulted in invalid payload.

harshavardhana:
This was experienced and reported by @elcolie.

### DIFF
--- a/minio/signer.py
+++ b/minio/signer.py
@@ -40,7 +40,7 @@ from .helpers import (ignore_headers, encode_to_hex,
 _SIGN_V4_ALGORITHM = 'AWS4-HMAC-SHA256'
 
 # Hardcoded S3 header value for X-Amz-Content-Sha256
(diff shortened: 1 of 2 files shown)


minio/signer.py
```diff
@@ -40,7 +40,7 @@
 _SIGN_V4_ALGORITHM = 'AWS4-HMAC-SHA256'
 
 # Hardcoded S3 header value for X-Amz-Content-Sha256
-_UNSIGNED_PAYLOAD = b'UNSIGNED-PAYLOAD'
+_UNSIGNED_PAYLOAD = 'UNSIGNED-PAYLOAD'
```
vadmeste: I've described the reason of declaring _UNSIGNED_PLAYLOAD as bytes in the PR description and not in the code (sorry!). This is a workaround to support both 2.7 and 3.x versions.

harshavardhana: It's not needed.

PR #430, opened 2016-12-01
Add List Objects V2 API (Fixes #376)

### LINKED ISSUE #376: api: Implement support for ListObjectsV2. supporting new S3 list version 2 type.

donatello:
* Adds new `list_objects_v2` function that uses the V2 List Objects API
* Adds unit test and functional test.
* Adds docs and example.

This patch also incorporates a better namespace aware XML parsing style. The existing `list_objects` function should also be separately improved to use this style.


donatello:
This patch is ready for review.

docs/API.md
```diff
@@ -227,6 +227,45 @@ for obj in objects:
...
+
+
+
+__Example__
+
+```py
+
+# List all object paths in bucket that begin with my-prefixname.
+objects = minioClient.list_objects_v2('mybucket', prefix='my-prefixname',
+                              recursive=True)
```
harshavardhana: README.md needs to be updated.. and linked here as well.

donatello: This is fixed.

PR #434, opened 2016-12-06
Add support for `response_headers` argument on `presigned_get_object`

g-cassie:
This allows users to specify additional querystring params supported by S3 API such as
`response-content-type` and `response-content-disposition`.

This was done following discussion on Gitter with @harshavardhana and @balamurugana

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -1286,7 +1286,8 @@ class Minio(object):
                 return
 
     def presigned_get_object(self, bucket_name, object_name,
-                             expires=timedelta(days=7)):
+                             expires=timedelta(days=7),
+                             response_headers=None):
         """
         Presigns a get object request and provides a url
 
@@ -1313,7 +1314,8 @@ class Minio(object):
 
         return self._presigned_get_partial_object(bucket_name,
                                                   object_name,
-                                                  expires)
+                                                  expires,
+                                                  response_headers=response_headers)
 
     def presigned_put_object(self, bucket_name, object_name,
                              expires=timedelta(days=7)):
@@ -1463,7 +1465,8 @@ class Minio(object):
 
     def _presigned_get_partial_object(self, bucket_name, object_name,
                                       expires=timedelta(days=7),
-                                      offset=0, length=0):
+                                      offset=0, length=0,
+                                      response_headers=None):
         """
         Presigns a get partial object request and provides a url,
         this is a internal function not exposed.
@@ -1491,6 +1494,7 @@ class Minio(object):
                              bucket_name=bucket_name,
                              object_name=object_name,
                              bucket_region=region)
+
         headers = {}
 
         if request_range:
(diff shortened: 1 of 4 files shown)


minio/api.py
```diff
@@ -1286,7 +1286,8 @@ def remove_incomplete_upload(self, bucket_name, object_name):
                 return
 
     def presigned_get_object(self, bucket_name, object_name,
-                             expires=timedelta(days=7)):
+                             expires=timedelta(days=7),
+                             response_headers=None):
```
balamurugana: 1. You could name the param as `custom_headers`
2. You could have pythonic way of emptiness like `response_headers={}`

minio/signer.py
```diff
@@ -96,6 +96,9 @@ def presign_v4(method, url, access_key, secret_key, region=None,
 
     headers_to_sign = dict(headers)
 
+    if response_headers is not None:
+        headers_to_sign.update(response_headers)
+
```
balamurugana: If `response_headers` is an empty `dict`, you could do simply `headers_to_sign.update(response_headers)`

minio/api.py
g-cassie: @balamurugana that would work here but due to dangers of passing a mutable as a default argument in python I think best practice is to pass `None`. See this SO for some discussion on that:
http://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

Happy to change it though - just let me know.

I'm open to changing to `custom_headers` but I chose `response_headers` because it matches aws docs (http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html).  Also the headers all follow the `repsonse-*` naming convention.

balamurugana:
> that would work here but due to dangers of passing a mutable as a default argument in python I think best practice is to pass None. See this SO for some discussion on that:
> http://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python
> 
> Happy to change it though - just let me know.
> 

As Python objects are mutable by default, programmers would need to do little care.  I am OK to keep it as `None`.  One minor advantage with `{}` is readability ie its understood by looking into the prototype over reading further docs.


> I'm open to changing to custom_headers but I chose response_headers because it matches aws docs (http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html). Also the headers all follow the repsonse-* naming convention.

:+1:

balamurugana:
:+1: LGTM

harshavardhana:
This is great work thanks @g-cassie - much appreciated.

g-cassie:
Thanks for quick review/merge @balamurugana and @harshavardhana 

Would it be possible to cut a release and publish to pypi?

harshavardhana:
> Would it be possible to cut a release and publish to pypi?

Pushed 2.0.4 @g-cassie

Issue #437, opened 2016-12-08
Need to review/refactor usage of str and bytes throughout lib to ensure python 3 functionality.


donatello:

harshavardhana:
Closing this as fixed from #445 and #442

PR #442, opened 2016-12-19
Part 1 in Strings/Bytes refactor:

donatello:
get_sha256(): Now accepts either `bytes` or `str` type (i.e. unicode)
and always returns hexadecimal encoded digest as a `str` (unicode)
type in both Python 3.x and Python 2.x. This greatly simplifies code
related to sha256 hash usage.

Towards resolving #437.

## Motivation

This refactor is to improve the interaction of string types in the library and improve the code quality.

### DIFF
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -34,7 +34,7 @@ import os
 import errno
 import math
 
-from .compat import urlsplit, urlencode
+from .compat import urlsplit, urlencode, str, bytes, basestring
 from .error import (InvalidBucketError, InvalidEndpointError,
                     InvalidArgumentError)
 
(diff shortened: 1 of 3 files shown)


donatello:
Both functional tests and unit tests are passing locally on my laptop.

minio/helpers.py
```diff
@@ -539,16 +539,24 @@ def encode_object_name(object_name):
 
 def get_sha256(content):
     """
-    Calculate sha256 digest of input byte array.
+    Calculate sha256 hexdigest of content.
 
-    :param content: Input byte array.
-    :return: sha256 digest of input byte array.
+    :param content: Input string or bytes.
```
krisis: Add a comment explaining how string values are encoded using `utf-8` encoding.

minio/helpers.py
```diff
@@ -479,7 +479,7 @@ def is_valid_bucket_notification_config(notifications):
                     'configuration'
                 )
             # check if 'Id' key is present, it should be a string
```
krisis: I think 'string' in this comment now includes both bytes and str (as imported from .compat). Could you update the comment to reflect it.

donatello: Done.

minio/helpers.py
donatello: Done.

donatello:
Updated the commit to rename the function and make clear its intent.

PR #443, opened 2016-12-19
XML Parsing refactor:

### LINKED ISSUE #432: Refactor list objects API support
Need to improve XML parsing style (using namespace aware parsing functionality) to simplify code.

donatello:
Use namespace aware functions throughout and eliminate usage of large
repetitive constants.

Fixes #432.

### DIFF
--- a/minio/parsers.py
+++ b/minio/parsers.py
@@ -50,15 +50,24 @@ else:
 _S3_NS = {'s3' : 'http://s3.amazonaws.com/doc/2006-03-01/'}
 
 
-def get_element_text(element, xpath, ns=_S3_NS):
-    """
-    Get element text for a given xpath.
+def get_element_text(element, xpath, strict=True, ns=_S3_NS):
+    """Get element text for a given xpath.
 
     :param element: Element xml object.
-    :param xpath: XML attribute inside the element.
+
+    :param xpath: XML xpath to search for descendant element. Not all
+    xpath syntax is supported - please refer to the library doc.
+
+    :param strict: If xpath is not resolvable, raises AttributeError
+    if true, otherwise returns None.
+
     :param ns: Necessary namespace to look for.
+
     """
-    return element.find(xpath, ns).text
+    if strict:
+        return element.find(xpath, ns).text
+    else:
+        return element.findtext(xpath, None, ns)
 
 def parse_multipart_upload_result(data):
     """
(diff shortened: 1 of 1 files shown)


minio/parsers.py
```diff
@@ -141,6 +141,39 @@ def parse_list_buckets(data):
         raise InvalidXMLError('Invalid XML provided for "ListBucketsResult" '
                               'some fields are missing. Message: {0}'.format(error.message))
 
+def _parse_objects_from_xml_elts(bucket_name, contents, common_prefixes):
+    """Internal function that extracts objects and common prefixes from
+    list_objects responses.
+    """
+    objects = []
+    for content in contents:
+        key_elt = content.find('s3:Key', _S3_NS)
```
balamurugana: Couple of thoughts.

1. cElementTree works with xpath ie `name = tree.find('volStatus/volumes/volume/volName').text`.

2. you could extend or have generic functions on top of  cElementTree classes to aware of namespace and returning urldecoded value by default.  You could avoid repeation of logic like `content.find("element", _S3_NS)`

minio/parsers.py
```diff
@@ -141,6 +141,39 @@ def parse_list_buckets(data):
...
+        object_name = urldecode(key_elt.text) if key_elt != None else None
```
balamurugana: You could tell me why we need to do take care of `None`? `urldecode(key_elt.text)` would throw `NoneType` exception which means the xml format is invalid.

donatello: I did the change from the point-of-view of only refactoring without changing behaviour - the earlier version did the same.

Let me add a commit here that throws exceptions if certain fields are not present, after taking a look at the spec once again.

minio/parsers.py
```diff
@@ -421,6 +380,8 @@ def _iso8601_to_localized_time(date_string):
     :param date_string: iso8601 formatted date string.
     :return: :class:`datetime.datetime`
     """
+    if date_string is None:
```
donatello: Removed this check now.

minio/parsers.py
donatello: PR is updated.

minio/parsers.py
donatello: This is changed now, and addresses these points as well.

donatello:
Closing in favour of #444

PR #445, opened 2016-12-20
Part 2 of Strings/Bytes Refactor:

donatello:
- Modify `get_md5` to `get_md5_base64digest`. This function now works
  for both str and bytes type inputs and returns str
  type (i.e. unicode string). It performs base64 encoding. This
  simplifies md5 related hashing code. It has the same behaviour in
  both Py2 and Py3 now.

- Add new Hasher class that provides `md5` an `sha256` hashers similar
  to hashlib, but always returns unicode string encoded hex/base64
  digests.

Towards resolving #437

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -35,7 +35,6 @@ from datetime import datetime, timedelta
 import io
 import json
 import os
-import hashlib
 import itertools
 
 # Dependencies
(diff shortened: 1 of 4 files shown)


donatello:
Pinging reviewers!

Issue #448, opened 2016-12-23
set_bucket_policy doesn't work


ChillarAnand:
In the examples, it is mentioned that [this should work](https://github.com/minio/minio-py/blob/eaca7a4279471151126e4b4dd3eaa5880550b7d5/examples/set_bucket_policy.py#L32):

    client.set_bucket_policy('my-bucketname', '', Policy.READ_ONLY)

I tried this on local system but it throws error
```py
import minio

mc = minio.Minio(
    endpoint='0.0.0.0:9000',
    access_key='foo',
    secret_key='1l+E',
    secure=False,
)

mc.set_bucket_policy('test-public', '', minio.policy.Policy.READ_ONLY)
```

```py
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/chillaranand/.01/python/minio_exp.py", line 12, in <module>
    mc.set_bucket_policy('mycroft-public', '', minio.policy.Policy.READ_ONLY)
  File "/home/chillaranand/.virtualenvs/sherlock/lib/python3.5/site-packages/minio/api.py", line 369, in set_bucket_policy
    bucket_name, prefix)
  File "/home/chillaranand/.virtualenvs/sherlock/lib/python3.5/site-packages/minio/policy.py", line 496, in set_policy
    bucket_name, prefix)
  File "/home/chillaranand/.virtualenvs/sherlock/lib/python3.5/site-packages/minio/policy.py", line 236, in _remove_statements
    in_use = _get_in_use_policy(statements, bucket_name, prefix)
  File "/home/chillaranand/.virtualenvs/sherlock/lib/python3.5/site-packages/minio/policy.py", line 135, in _get_in_use_policy
    resource.startswith(resource_prefix)):
AttributeError: 'list' object has no attribute 'startswith'
```

But the same works with `mc`

    → mc policy download local/mycroft-public
    Access permission for ‘local/mycroft-public’ is set to ‘download’

harshavardhana:
Fixed by #449

PR #449, opened 2016-12-23
Update resource as per new policy

ChillarAnand:
minio is returning resource as a list

```
{"Version":"2012-10-17","Statement":[{"Action":["s3:GetBucketLocation","s3:ListBucketMultipartUploads"],"Effect":"Allow","Principal":{"AWS":["*"]},"Resource":["arn:aws:s3:::mycroft-public"],"Sid":""},{"Action":["s3:GetBucketLocation"],"Effect":"Allow","Principal":{"AWS":["*"]},"Resource":["arn:aws:s3:::mycroft-public"],"Sid":""},{"Action":["s3:AbortMultipartUpload","s3:DeleteObject","s3:ListMultipartUploadParts","s3:PutObject"],"Effect":"Allow","Principal":{"AWS":["*"]},"Resource":["arn:aws:s3:::mycroft-public/*"],"Sid":""},{"Action":["s3:GetBucketLocation"],"Effect":"Allow","Principal":{"AWS":"*"},"Resource":["arn:aws:s3:::mycroft-public"],"Sid":""},{"Action":["s3:ListBucket"],"Effect":"Allow","Principal":{"AWS":"*"},"Resource":["arn:aws:s3:::mycroft-public"],"Sid":""},{"Action":["s3:GetObject"],"Effect":"Allow","Principal":{"AWS":"*"},"Resource":["arn:aws:s3:::mycroft-public/*"],"Sid":""},{"Action":["s3:GetObject"],"Effect":"Allow","Principal":{"AWS":"*"},"Resource":["arn:aws:s3:::mycroft-public/*"],"Sid":""}]}
```

### DIFF
--- a/minio/policy.py
+++ b/minio/policy.py
@@ -53,8 +53,20 @@ class Policy:
     WRITE_ONLY = "writeonly"
 
 
-# Returns 'Action' value as list always.
+def _get_resource(statement):
+    """
+    Always return resource as a list
+    """
+    resource = statement.get('Resource')
+    if isinstance(resource, basestring):
+        resource = [resource]
+    return resource
+
+
 def _get_action(statement):
+    """
+    Always return action as a list
+    """
     action = statement.get('Action', [])
     if isinstance(action, basestring):
         action = [action]
@@ -119,6 +131,13 @@ def _new_statements(policy, bucket_name, prefix=''):
             _new_object_statement(policy, bucket_name, prefix))
 
 
+def _filter_resources(prefix, resources):
+    """
+    Returns a list of resources, which starts with given prefix
+    """
+    return list(filter(lambda r: r.startswith(prefix), resources))
+
+
 # Returns whether given bucket statements are used by other than given
 # prefix statements.
 def _get_in_use_policy(statements, bucket_name, prefix=''):
@@ -129,10 +148,10 @@ def _get_in_use_policy(statements, bucket_name, prefix=''):
               Policy.WRITE_ONLY: False}
 
     for s in statements:
-        resource = s['Resource']
+        resource = _get_resource(s)
         actions = set(_get_action(s))
         if (resource != object_resource and
-                resource.startswith(resource_prefix)):
+                _filter_resources(resource_prefix, resource)):
             if (not in_use[Policy.READ_ONLY] and
                     actions & _READ_ONLY_OBJECT_ACTIONS):
                 in_use[Policy.READ_ONLY] = True
@@ -278,7 +297,7 @@ def _remove_statements(statements, policy, bucket_name, prefix=''):
     resource_prefix = _AWS_RESOURCE_PREFIX + bucket_name + '/'
     for s in out:
         resource = s['Resource']
-        if (resource.startswith(resource_prefix) and
+        if (resource_startswith(resource_prefix, resource) and
                 resource not in s3_prefix_value):
             skip_bucket_statement = False
             break


donatello:
@balamurugana Could you review this? Also related to https://github.com/minio/minio-py/issues/448

ChillarAnand:
This fixes https://github.com/minio/minio-py/issues/448

minio/policy.py
```diff
@@ -132,7 +132,7 @@ def _get_in_use_policy(statements, bucket_name, prefix=''):
         resource = s['Resource']
         actions = set(_get_action(s))
         if (resource != object_resource and
-                resource.startswith(resource_prefix)):
+                resource[0].startswith(resource_prefix)):
```
balamurugana: 1. As per Amazon AWS dynamic JSON, `Resource` can be a string or a list of string.  Assuming as a list would break logically.  The right fix would be type checking like
```py
resource = s['resource']
if isinstance(resource, basestring):
    resource = [resource]
```

2. We shouldn't assume `resource[0]` instead it could match in any item in the list.

ChillarAnand: @balamurugana Thanks. Fixed it.

minio/policy.py
```diff
@@ -66,7 +81,7 @@ def _new_bucket_statement(policy, bucket_name, prefix=''):
     if policy == Policy.NONE:
         return []
 
-    bucket_resource = _AWS_RESOURCE_PREFIX + bucket_name
+    bucket_resource = [_AWS_RESOURCE_PREFIX + bucket_name]
```
balamurugana: Why do you want `bucket_resource` as list?

minio/policy.py
```diff
@@ -129,10 +148,10 @@ def _get_in_use_policy(statements, bucket_name, prefix=''):
               Policy.WRITE_ONLY: False}
 
     for s in statements:
-        resource = s['Resource']
+        resource = _get_resource(s)
```
balamurugana: if `s['Resource']` would raise an exception which denotes an wrong JSON.  This approach swallows the error.

minio/policy.py
```diff
@@ -119,6 +134,10 @@ def _new_statements(policy, bucket_name, prefix=''):
             _new_object_statement(policy, bucket_name, prefix))
 
 
+def resource_startswith(prefix, resources):
+    return any([resource.startswith(prefix) for resource in resources])
```
balamurugana: is it good idea to use `filter()` function something like 
```py
return filter(lambda r: r.startswith(prefix), resources)
```

I am not sure which one is better.

minio/policy.py
ChillarAnand: Removed default value. Thanks.

minio/policy.py
ChillarAnand: I thought resource has to be a list. Didn't know, it can either be list or string.

minio/policy.py
balamurugana: :+1:

subho007:
LGTM

minio/policy.py
```diff
@@ -119,6 +131,10 @@ def _new_statements(policy, bucket_name, prefix=''):
             _new_object_statement(policy, bucket_name, prefix))
 
 
+def resource_startswith(prefix, resources):
+    return list(filter(lambda r: r.startswith(prefix), resources))
```
balamurugana: 1. filter() always returns list by default. wrapping filter() return is unnecessary.
2. rename resource_startswith() to _filter_resources() and add a comment what this function does

ChillarAnand:
```
In [51]: bool(filter(lambda r: r.startswith('c'), ['aa', 'bb']))
Out[51]: True
```
In python 3, filter returns filter object and its bool is always true. So we have to convert to list.

Documented `_filter_resources`. Thanks.

balamurugana: :+1:  `filter()` returns an iterable in py3.

ChillarAnand:
Can you make a new release?

harshavardhana:
@ChillarAnand will do soon.

PR #451, opened 2016-12-26
add pyflakes check

balamurugana:


minio/api.py
```diff
@@ -375,10 +375,6 @@ def set_bucket_policy(self, bucket_name, prefix, policy_access):
             policy_dict['Statement'] = statements
             content = json.dumps(policy_dict)
 
-            headers = {
-                'Content-Length': str(len(content)),
-                'Content-Md5': get_md5_base64digest(content)
```
harshavardhana: why are you removing this?

balamurugana: It is not used.  pyflakes error.

harshavardhana: we should set it.. md5sum validation is needed for the policy json sent.

PR #465, opened 2017-01-30
Make HTTP(S) responses always preload content by default:

### LINKED ISSUE #458: Connections to service are not being re-used - starting new HTTP connection for every request.
Hi!

Looking at the code, the minio client uses HTTP pooling. Still, each time I issue a `client.get_object()` or `client.put_object()`, I get the following log message: `Starting new HTTP connection (34): minio`, with the number increasing each time.

I use the same `minio.Minio` object for all requests, so the host and port are the same each time.

```
$ pip freeze | grep minio
minio==2.0.6
```

donatello:
For most API requests (except GET object), the response is not large
and can be loaded into memory. Preloading content makes re-use of
connections automatic - the connection is returned to the connection
pool before the response object is returned to the caller of
urlopen().

The only exception, GET object, uses preload_content=False - this
allows the caller to incrementally process the response - but as a
caveat would need to release the connection manually by calling
`release_conn()` on the returned object. Docstrings have been
updated to reflect this.

Fixes #458.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -222,8 +222,7 @@ class Minio(object):
 
         response = self._http.urlopen(method, url,
                                       body=content,
-                                      headers=headers,
-                                      preload_content=False)
+                                      headers=headers)
 
         if response.status != 200:
             response_error = ResponseError(response)
@@ -259,8 +258,7 @@ class Minio(object):
 
         response = self._http.urlopen(method, url,
                                       body=None,
-                                      headers=headers,
-                                      preload_content=False)
+                                      headers=headers)
 
         if self._trace_output_stream:
             dump_http(method, url, headers, response,
@@ -685,6 +683,9 @@ class Minio(object):
                 file_part_data.write(data)
                 total_written += len(data)
 
+            # Release the connection from the response at this point.
+            response.release_conn()
+
             # Verify if we wrote data properly.
             if total_written < content_size:
                 msg = 'Data written {0} bytes is smaller than the' \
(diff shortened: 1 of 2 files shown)


harshavardhana:
In this can we make release_conn() to be `close()` api basically treating get_object returned as an FD.
and then a user closes either with explicit `close()` .  Extra polating from this someone can do 

```
with client.get_object('my-bucketname', 'my-objectname') as obj_reader:
```

and close be handled automatically? - `release_conn()` is kind of non-standard and hard to follow. I wonder why `urllib3` cannot do this ideally. 

Thoughts?

donatello:
@harshavardhana 

> In this can we make release_conn() to be close() api basically treating get_object returned as an FD.
> and then a user closes either with explicit close() . Extra polating from this someone can do
> 
> with client.get_object('my-bucketname', 'my-objectname') as obj_reader:
> 
> and close be handled automatically? - release_conn() is kind of non-standard and hard to follow. I wonder why urllib3 cannot do this ideally.
> 
> Thoughts?

Turns out this is not worth the trouble - the [response module](https://urllib3.readthedocs.io/en/latest/reference/index.html#module-urllib3.response) is already an instance io.IOBase, and supports with `with` statement with the meaning that it will call `close()` on the response object - the [close()](https://github.com/shazow/urllib3/blob/master/urllib3/response.py#L474) method closes the connection without returning it to the pool. 

Calling `release_conn()` is not strictly necessary unless the user is particular about re-using the connection - otherwise there is no leakage of resources - when the object is garbage collected, the connection is simply closed.

PR #474, opened 2017-02-11
Allow bucket region cache to be overriden.

harshavardhana:
Add region= option to our API constructor. This option
is added to avoid making calls to server when an
application written using `minio-py` is only meant
to generate presigned URLs which is a quite common use
case.

Setting this option - overrides any future region
changes and makes the API more stricter. For example
if you are going to use your initialized client
against a bucket which exists in a different region
all API operations would fail.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -60,8 +60,9 @@ __Parameters__
 |:---|:---|:---|
 | `endpoint`  | _string_  | S3 object storage endpoint.  |
 | `access_key`  | _string_  | Access key for the object storage endpoint. (Optional if you need anonymous access).  |
-|  `secret_key` | _string_  |  Secret key for the object storage endpoint. (Optional if you need anonymous access). |
+| `secret_key` | _string_  |  Secret key for the object storage endpoint. (Optional if you need anonymous access). |
 | `secure`  |_bool_   | Set this value to `True` to enable secure (HTTPS) access. (Optional defaults to `True`).  |
+| `region`  |_string_ | Set this value to override automatic bucket location discovery. (Optional defaults to `None`). |
 
 __Example__
 
--- a/minio/api.py
+++ b/minio/api.py
@@ -100,15 +100,22 @@ class Minio(object):
         client = Minio('play.minio.io:9000')
         client = Minio('s3.amazonaws.com', 'ACCESS_KEY', 'SECRET_KEY')
 
+        # To override auto bucket location discovery.
+        client = Minio('play.minio.io:9000', 'ACCESS_KEY', 'SECRET_KEY',
+                       region='us-east-1')
+
     :param endpoint: Hostname of the cloud storage server.
     :param access_key: Access key to sign self._http.request with.
     :param secret_key: Secret key to sign self._http.request with.
     :param secure: Set this value if wish to make secure requests.
          Default is True.
+    :param region: Set this value to override automatic bucket
+         location discovery.
     :return: :class:`Minio <Minio>` object
     """
     def __init__(self, endpoint, access_key=None,
-                 secret_key=None, secure=True):
+                 secret_key=None, secure=True,
+                 region=None):
         # Validate endpoint.
         is_valid_endpoint(endpoint)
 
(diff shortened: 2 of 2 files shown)


minio/api.py
```diff
@@ -193,6 +196,14 @@ def make_bucket(self, bucket_name, location='us-east-1'):
         """
         is_valid_bucket_name(bucket_name)
 
+        ## Region already set in constructor, validate if
+        ## caller requested bucket location is same.
+        if self._region:
+            if self._region != location:
+                raise InvalidArgumentError("Configured region {0}, requested"
```
donatello: This doesnt seem right - how does one make a bucket in a region other than `self._region`? It seems like it is needed to reset the Minio Client object to do this.

minio/api.py
```diff
@@ -1695,13 +1706,20 @@ def _get_bucket_region(self, bucket_name):
         :param bucket_name: Bucket name for which region will be fetched.
         :return: Region of bucket name.
         """
+
+        # Region set in constructor, return right here.
+        if self._region:
+            return self._region
```
donatello: This is true only for buckets created with the same Minio instance - not necessary for other buckets.

minio/api.py
harshavardhana: If the region is provided one cannot.

minio/api.py
harshavardhana: region is provided to override all the buckets and all calls.. There is no cache based approach if region is set.

harshavardhana:
NOTE: This option is provided for special use cases where users of presigned URL wish to not make a network call i.e avoid GetBucketLocation and have all their buckets in a well defined region. In a scenario like this override is useful.

balamurugana:
> NOTE: This option is provided for special use cases where users of presigned URL wish to not make a network call i.e avoid GetBucketLocation and have all their buckets in a well defined region. In a scenario like this override is useful.

If this is the only use case, why can't we have presigned url calls accept region as parameter?

docs/API.md
```diff
@@ -60,8 +60,9 @@ __Parameters__
 |:---|:---|:---|
 | `endpoint`  | _string_  | S3 object storage endpoint.  |
 | `access_key`  | _string_  | Access key for the object storage endpoint. (Optional if you need anonymous access).  |
-|  `secret_key` | _string_  |  Secret key for the object storage endpoint. (Optional if you need anonymous access). |
+| `secret_key` | _string_  |  Secret key for the object storage endpoint. (Optional if you need anonymous access). |
 | `secure`  |_bool_   | Set this value to `True` to enable secure (HTTPS) access. (Optional defaults to `True`).  |
+| `region`  |_string_ | Set this value to override bucket region cache. |
```
donatello: There is not much elaboration on what "region cache" is - so for a user of the lib this may sound new. Better "Set this value to explicitly restrict all requests to a particular region - it disables region discovery and caching behavior."

minio/api.py
donatello: Alternatively, instead of doing something that appears to change behaviour a lot, we can get equivalent behaviour by modifying the pre-sign APIs to take an explicit region argument.

harshavardhana:
> If this is the only use case, why can't we have presigned url calls accept region as parameter?

We shouldn't that is not right API style region is just like access and secret keys should be passed along with the constructor. Presigned URL doesn't require that as a parameter and users are not going to be expecting it to be passed along as an option.

docs/API.md
harshavardhana: Done

harshavardhana:
> Alternatively, instead of doing something that appears to change behaviour a lot, we can get equivalent behaviour by modifying the pre-sign APIs to take an explicit region argument.

For the most part users are usually accessing the buckets in the same region, creating a bucket is a rare operation that to in regions other than their current region.   Adding such an option in presignURL causes confusion where an active cache of a bucket region - which could be different than the one user just supplied leading to unexpected bugs. 

A constructor is the rightful place for this along with access and secret keys, allows for the behavior to be predictable and consistent - when region is provided v/s otherwise.

PR #484, opened 2017-02-23
Return consistent outputs for put_object/fput_object

### LINKED ISSUE #480: api: return stat results from fget_object
HashBackup stores the etag from a put operation (usually md5) and verifies it on a get operation.  fget_object already does a stat_object before the get, so it would be easy to return from fget_object.  I can do a PR for that if acceptable.

### LINKED ISSUE #481: doc: fput_object returns either etag or MultipartUploadResult
fput_object already returns either etag (smallish files) or MultipartUploadResult, but it isn't documented to return anything.  It would be useful to document the return value so that etag or mur.etag could be stored and verified for a future get operation.  Want a PR for that?

harshavardhana:
Additionally also return the stat info of the original
object for fget_object.

Fixes #480
Fixes #481

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -215,7 +215,6 @@ __Return Value__
 |``object.last_modified`` |_datetime.datetime_ | modified time stamp.  |
 
 
-
 __Example__
 
 ```py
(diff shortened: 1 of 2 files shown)


docs/API.md
```diff
@@ -783,27 +801,33 @@ __Parameters__
...
+|``etag``|_string_  |Object etag computed by the server.  |
+
 __Example__
 
 The maximum size of a single object is limited to 5TB. fput_object transparently uploads objects larger than 5MiB in multiple parts. This allows failed uploads to resume safely by only uploading the missing parts. Uploaded data is carefully verified using MD5SUM signatures.
 
 ```py
 
-# Put an object 'myobject' with contents from '/tmp/otherobject'.
+# Put an object 'myobject' with contents from '/tmp/otherobject', prints hex md5sum of the object if uploaded successfully.
```
donatello: This is not accurate - for multipart objects the etag != hex m5sum

docs/API.md
```diff
@@ -744,18 +756,24 @@ __Parameters__
...
+
 __Example__
 
 The maximum size of a single object is limited to 5TB. put_object transparently uploads objects larger than 5MiB in multiple parts. This allows failed uploads to resume safely by only uploading the missing parts. Uploaded data is carefully verified using MD5SUM signatures.
 
 ```py
 
 import os
-# Put a file with default content-type.
+# Put a file with default content-type, upon success prints hex md5sum of the object.
```
donatello: This is not accurate - for multipart objects the etag != hex m5sum

harshavardhana: Oops yes.

PR #494, opened 2017-03-18
Custom certificate bundles

MrRedAmber:
Added a parameter to the Minio class initialization method that allows users to provide their own certificate bundle. One example use case scenario would be if a user wanted to provide a self signed certificate to use with their Minio object. If the user doesn't provide a certificate bundle themselves, Minio will default to using Certifi's (certifi.where()) certificate bundle.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -119,7 +119,8 @@ class Minio(object):
     def __init__(self, endpoint, access_key=None,
                  secret_key=None, secure=True,
                  region=None,
-                 timeout=None):
+                 timeout=None,
+				 certificate_bundle=certifi.where()):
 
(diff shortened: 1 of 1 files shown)


minio/api.py
```diff
@@ -119,7 +119,8 @@ class Minio(object):
     def __init__(self, endpoint, access_key=None,
                  secret_key=None, secure=True,
                  region=None,
-                 timeout=None):
+                 timeout=None,
+				 certificate_bundle=certifi.where()):
```
harshavardhana: tabbing.. ?

MrRedAmber: Look at latest commit to branch

PR #505, opened 2017-04-16
api: Return custom metadata in stat_object.

harshavardhana:
We allow custom metadata to be set
in fput_object(). So we shall properly
return custom metadata back as requested
by the caller.

Fixes https://github.com/minio/minio/issues/4136

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -645,6 +645,7 @@ __Return Value__
 |``obj.etag``|_string_|etag of the object.|
 |``obj.content_type``|_string_  | Content-Type of the object.|
 |``obj.last_modified``|_time.time_  |modified time stamp.|
+|``obj.metadata`` |_dict_ | Contains any additional metadata on the object. |
 
 __Example__
 
@@ -816,6 +817,7 @@ __Return Value__
 |``obj.etag``|_string_|etag of the object.|
 |``obj.content_type``|_string_  | Content-Type of the object.|
 |``obj.last_modified``|_time.time_  |modified time stamp.|
+|``obj.metadata`` |_dict_ | Contains any additional metadata on the object. |
 
 
 __Example__
--- a/minio/api.py
+++ b/minio/api.py
@@ -1048,11 +1048,26 @@ class Minio(object):
         size = int(response.headers.get('content-length', '0'))
         content_type = response.headers.get('content-type', '')
         last_modified = response.headers.get('last-modified')
+
+        ## Supported headers for object.
+        supported_headers = [
+            'cache-control',
+            'content-encoding',
+            'content-disposition',
+            ## Add more supported headers here.
+        ]
+
+        ## Capture only custom metadata.
+        custom_metadata = dict()
+        for k in response.headers:
+            if k in supported_headers or k.lower().startswith('x-amz-meta-'):
+                custom_metadata[k] = response.headers.get(k)
+
         if last_modified:
             http_time_format = "%a, %d %b %Y %H:%M:%S GMT"
             last_modified = mktime(strptime(last_modified, http_time_format))
-        return Object(bucket_name, object_name, content_type=content_type,
-                      last_modified=last_modified, etag=etag, size=size)
+        return Object(bucket_name, object_name, last_modified, etag, size,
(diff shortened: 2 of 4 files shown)


minio/api.py
```diff
@@ -1045,14 +1045,35 @@ def stat_object(self, bucket_name, object_name):
...
+            'cache-control',
+            'content-encoding',
+            'content-disposition',
+            ## Add more supported headers here.
+        ]
+
+        ## Capture only custom metadata.
+        custom_metadata = dict()
+        for k in response.headers:
+            if k in supported_headers:
```
krisis: We can combine the two conditions as,
```python
if k in supported_headers or k.lower().startswith('x-amz-meta'):
    custom_metadata[k] = response.headers.get(k)
```

harshavardhana: sure @krisis good idea.

harshavardhana:
PR is updated.

minio/api.py
```diff
@@ -1045,14 +1045,34 @@ def stat_object(self, bucket_name, object_name):
...
+            'cache-control',
+            'content-encoding',
+            'content-disposition',
+            ## Add more supported headers here.
+        ]
+
+        ## Capture only custom metadata.
+        custom_metadata = dict()
+        for k in response.headers:
+            if (k in supported_headers or
```
krisis: what is the need to paranthesise the expression in if-block?

harshavardhana: Indentation

harshavardhana: Remove the indentation block.

tests/functional/tests.py
```diff
@@ -133,6 +133,13 @@ def main():
     # Get a full object locally.
     client.fget_object(bucket_name, object_name, 'newfile-f')
 
+    client.fput_object(bucket_name, object_name+'-f', 'testfile',
+                       metadata={'x-amz-meta-testing': 'value'})
+
+    stat = client.fget_object(bucket_name, object_name+'-f', 'newfile-f-custom')
+    if not stat.metadata.has_key('X-Amz-Meta-Testing'):
```
vadmeste: Can we also check the value of this header ?

docs/API.md
```diff
@@ -645,6 +645,7 @@ __Return Value__
 |``obj.etag``|_string_|etag of the object.|
 |``obj.content_type``|_string_  | Content-Type of the object.|
 |``obj.last_modified``|_time.time_  |modified time stamp.|
+|``obj.metadata`` |_dict_ | Contains any additional metadata on the object. |
```
vadmeste: Will we bump a new major version ?

minio/api.py
```diff
@@ -1045,14 +1045,33 @@ def stat_object(self, bucket_name, object_name):
...
+            'cache-control',
+            'content-encoding',
+            'content-disposition',
+            ## Add more supported headers here.
+        ]
+
+        ## Capture only custom metadata.
+        custom_metadata = dict()
+        for k in response.headers:
+            if k in supported_headers or k.lower().startswith('x-amz-meta'):
```
harshavardhana: Sure

docs/API.md
harshavardhana: No we don't need to bump major version the type doesn't change neither it breaks anything. It is just another field added.

vadmeste: oups yes.

tests/functional/tests.py
harshavardhana: Fixed

minio/api.py
```diff
@@ -1045,14 +1045,33 @@ def stat_object(self, bucket_name, object_name):
                                   object_name=object_name)
 
         etag = response.headers.get('etag', '').replace('"', '')
+        response.headers.pop('etag');
```
vadmeste: One last comment, not sure what are you doing here? removing header ? for what reason?

harshavardhana: Actually the idea was earlier to just save the response.headers verbatim. But since we are filtering now its not needed.

harshavardhana: Fixed good catch @vadmeste

PR #506, opened 2017-04-18
correct bucket_name to bucket.name

dwo:
I had written something thinking it should be 
```
buckets = client.list_buckets()
bucket_set = set([bucket.bucket_name for bucket in buckets])
```

but I actually had to write
```
buckets = client.list_buckets()
bucket_set = set([bucket.name for bucket in buckets])
```


dwo:
sorry, probably should have [ci skip]'d this 🙊

harshavardhana:
Thanks for the PR @dwo

PR #507, opened 2017-04-20
[WIP] api: Use credentials as authentication mechanism

### LINKED ISSUE #502: read ~/.mc/config.json consistently in all libraries
I'm building tools and libraries that allow people to use Minio/S3 storage with deep learning frameworks. This means that users are going to switch freely between writing Python code, Go code, and command line tools. This involves a lot of end user programming and scripting.

It would be nice if these different tools could share a common authentication infrastructure. In particular, it would be convenient if minio-py had an option of taking its keys and path shortcuts from ~/.mc/config.json

Of course, I can write a wrapper library, but having the option of having consistent paths and key management between the different Minio libraries and command line tools would seem to be a nice feature to have.

vadmeste:
This is very early to be review and it gives an idea about the design

Fixes #502

### DIFF
--- a/examples/list_buckets.py
+++ b/examples/list_buckets.py
@@ -16,11 +16,13 @@
 # Note: YOUR-ACCESSKEYID and YOUR-SECRETACCESSKEY are
 # dummy values, please replace them with original values.
 
-from minio import Minio
+import sys
+sys.path.append('/home/vadmeste/work/python/minio-py/')
 
-client = Minio('s3.amazonaws.com',
-               access_key='YOUR-ACCESSKEYID',
-               secret_key='YOUR-SECRETACCESSKEY')
+from minio import Minio
+from minio.credentials import iam_aws_credentials 
+
+client = Minio('localhost:9000', credentials=iam_aws_credentials(role_name="RoleName"), secure=False)
 
 buckets = client.list_buckets()
 
--- a/minio/credentials.py
+++ b/minio/credentials.py
@@ -1,0 +1,256 @@
+# -*- coding: utf-8 -*-
+# Minio Python Library for Amazon S3 Compatible Cloud Storage, (C) 2016, 2017 Minio, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


vadmeste:
@harshavardhana, can you comment on the code design and API before I start writing tests and documentation of this change ?

Basically, now we can do this now:
`client = Minio("localhost:9000", "access_key", "secret_key")`
or 
`client = Minio('localhost:9000', credentials=iam_aws_credentials(role_name="RoleName"))`
or
`client = Minio("localhost:9000", credentials=env_minio_credentials())`

etc..

examples/list_buckets.py
```diff
@@ -16,11 +16,13 @@
...
+import sys
+sys.path.append('/home/vadmeste/work/python/minio-py/')
+
 from minio import Minio
+from minio.credentials import iam_aws_credentials 
 
-client = Minio('s3.amazonaws.com',
-               access_key='YOUR-ACCESSKEYID',
-               secret_key='YOUR-SECRETACCESSKEY')
+client = Minio('localhost:9000', credentials=iam_aws_credentials(role_name="RoleName"), secure=False)
```
MrRedAmber: 80 character limit

minio/credentials.py
```diff
@@ -0,0 +1,256 @@
...
+from abc import ABCMeta, abstractmethod
+
+import os, json, urllib3, datetime
+import ConfigParser
+
+from .error import MinioError
+
+
+# credentials class
+class credentials(object):
```
MrRedAmber: Properly name all of your classes
```python
class Credentials(object):
```

minio/credentials.py
```diff
@@ -0,0 +1,256 @@
...
+        self._force_refresh = True
+
+# static_credentials
+class static_credentials(credentials_provider):
+
+    def __init__(self, access_key, secret_key, session_token=""):
+        super(static_credentials, self).__init__()
+        self._access_key = access_key
+        self._secret_key = secret_key
+        if session_token == None:
```
MrRedAmber: `session_token` will never be `None`, unless you strictly specify it

 if it is `None` then you shouldn't supply it to a method call

Just have...
```python
_session_token = session_token
```

vadmeste:
@harshavardhana, this is my first long PR for minio-py, can you review this quickly before I do tests for this PR ?

I didn't find any indication about the source format (80 cols width?) and naming classes and variables.

deekoder:
closing as per discussions

vadmeste:
This is closed because it needs to rewritten to follow minio-go style.

PR #515, opened 2017-05-03
Correct attribute initialization in ResponseError

fazpu:


minio/error.py
```diff
@@ -133,8 +133,6 @@ def __init__(self, response, method, bucket_name=None,
         self.object_name = object_name
```
fazpu: The attributes are already initialized here

riquito:
I've been bitten by this today, it took a while to understand what was going on. Any change that a release is near?

PR #518, opened 2017-05-08
Allow customizing urllib3 pool size

messense:
The default value `10` is not always fine, I got a lot of `connection poll is full` warnings in production.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -122,7 +122,8 @@ class Minio(object):
                  secret_key=None, secure=True,
                  region=None,
                  timeout=None,
-                 certificate_bundle=certifi.where()):
+                 certificate_bundle=certifi.where(),
+                 num_pools=10):
 
         # Validate endpoint.
         is_valid_endpoint(endpoint)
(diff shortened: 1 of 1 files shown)


harshavardhana:
Do not understand this. What is the issue does this solve @messense ?

messense:
You will get a lot of `Connection pool is full` warning when connections exceeds 10, it should be configurable.

```
2017-05-08 22:00:42,173 WARNING urllib3.connectionpool._put_conn:276   Connection pool is full, discarding connection: brm-prerender-cache.s3-ap-northeast-1.amazonaws.com
2017-05-08 22:00:42,208 WARNING urllib3.connectionpool._put_conn:276   Connection pool is full, discarding connection: brm-prerender-cache.s3-ap-northeast-1.amazonaws.com
```

messense:
API ref: http://urllib3.readthedocs.io/en/1.2.1/managers.html#urllib3.poolmanager.PoolManager

messense:
Actually, the whole connection pool manager should be configurable. In a multithreaded environment, it would be useful to have a connection pool with some connection cached instead of create new one every time.

PR #528, opened 2017-05-29
signv4: Strip port 80 & 443 from url

### LINKED ISSUE #527: mc presigned urls don't work if port is specified
"/home/minio/mc share download --expire=30m mymini/bucket/filename"

generates wrong presigned url if you use port number while configuring mymini

minio config should work with(out) port number specified in endpoint

vadmeste:
curl & browsers strip port 80 and 443 from host header when a user tries
to open a presigned urls, the thing that leads to a signature mismatch error.
This PR strips port 80/443 by default.

Fixes #527


harshavardhana:
We need to also check this across all libraries

PR #530, opened 2017-06-06
url: resource and query encoding should support unicode

### LINKED ISSUE #529: Unicode object name on python 2.7 causes exception
Trying to make django-minio-storage work with Python 2.7
In current active fork I have pull request here
https://github.com/py-pa/django-minio-storage/pull/15

In standard test with Unicode object name I get exception.
After some digging (and temporary code changes) for finding 
what is original exception I found this call stack:
```
__________ UploadTests.test_upload_and_get_back_file_with_funky_name ___________
tests/test_app/tests/upload_tests.py:75: in test_upload_and_get_back_file_with_funky_name
    self.media_storage.save("áčďěščřžýŽŇůúť.txt", ContentFile(b"12345"))
.tox/py27/lib/python2.7/site-packages/django/core/files/storage.py:53: in save
    name = self.get_available_name(name, max_length=max_length)
.tox/py27/lib/python2.7/site-packages/django/core/files/storage.py:77: in get_available_name
    while self.exists(name) or (max_length and len(name) > max_length):
minio_storage/storage.py:108: in exists
    self.bucket_name, self._sanitize_path(name))
.tox/py27/lib/python2.7/site-packages/minio/api.py:1045: in stat_object
    object_name=object_name)
.tox/py27/lib/python2.7/site-packages/minio/api.py:1861: in _url_open
    query=query)
.tox/py27/lib/python2.7/site-packages/minio/helpers.py:213: in get_target_url
    object_name = encode_object_name(object_name)
.tox/py27/lib/python2.7/site-packages/minio/helpers.py:481: in encode_object_name
    return urlencode(object_name)
/opt/python/2.7.12/lib/python2.7/urllib.py:61: in pathname2url
    return quote(pathname)
/opt/python/2.7.12/lib/python2.7/urllib.py:1299: in quote
    return ''.join(map(quoter, s))
E   KeyError: u'\xe1'
```

If you can see. We use urllib on python 2.7 and it is no unicode-ish.
We need to do things via urllib3 to support unicode.
Can you guys fix it?
My failing build result is here if that can be helpful:
https://travis-ci.org/py-pa/django-minio-storage/jobs/239789339

harshavardhana:
python2.7 lacks proper unicode support, so we need
a proper wrapper function for url encoding. With
this the url and query encodings works for all
unicode characters.

Fixes #529

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -45,7 +45,8 @@ import certifi
 
 # Internal imports
 from . import __title__, __version__
-from .compat import urlsplit, range, urlencode, basestring
+from .compat import (urlsplit, queryencode,
+                     range, basestring)
 from .error import (KnownResponseError, ResponseError, NoSuchBucket,
                     InvalidArgumentError, InvalidSizeError, NoSuchBucketPolicy)
 from .definitions import Object, UploadPart
@@ -841,7 +842,7 @@ class Minio(object):
         if conditions:
             headers = {k: v for k, v in conditions.items()}
 
-        headers['X-Amz-Copy-Source'] = urlencode(object_source)
+        headers['X-Amz-Copy-Source'] = queryencode(object_source)
         response = self._url_open('PUT',
                                   bucket_name=bucket_name,
                                   object_name=object_name,
@@ -946,8 +947,9 @@ class Minio(object):
 
         # Initialize query parameters.
         query = {
-            'max-keys': 1000
+            'max-keys': '1000'
         }
+
         # Add if prefix present.
         if prefix:
             query['prefix'] = prefix
@@ -1012,7 +1014,7 @@ class Minio(object):
 
         # Initialize query parameters.
         query = {
-            'list-type': 2
+            'list-type': '2'
         }
         # Add if prefix present.
         if prefix:
@@ -1228,7 +1230,7 @@ class Minio(object):
         # Initialize query parameters.
         query = {
             'uploads': '',
-            'max-uploads': 1000
+            'max-uploads': '1000'
         }
 
         if prefix:
@@ -1286,14 +1288,14 @@ class Minio(object):
 
         query = {
             'uploadId': upload_id,
-            'max-parts': 1000
+            'max-parts': '1000'
         }
 
         is_truncated = True
         part_number_marker = None
         while is_truncated:
             if part_number_marker:
-                query['part-number-marker'] = part_number_marker
+                query['part-number-marker'] = str(part_number_marker)
 
             response = self._url_open('GET',
                                       bucket_name=bucket_name,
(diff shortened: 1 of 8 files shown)


minio/api.py
```diff
@@ -44,7 +44,8 @@
 
 # Internal imports
 from . import __title__, __version__
-from .compat import urlsplit, range, urlencode, basestring
+from .compat import (urlsplit, queryencode,
```
Yurmusp: I see issue in your files/code.
It depends what you target for. I see it from the standpoint what I would target for.

I would target for have everything in your lib unicode. I mean all coming and outgoing strings.
For that typically programmer needs to have two things enabled shown as two lines below.

# encoding: utf-8
from __future__ import unicode_literals

First line says that this file is UTF-8.
Second line makes every string  to be unicode string in this file (not just base string with UTF-8 encoding in it).
What you have now is only first line out of two in this file.
So, your file has all base strings with UTF-8. NOT really unicode strings everywhere.
I'm not sure what you want.
You still can try to go with base strings and UTF-8, but that seems assumed to be more complicated way
 with many side effects that needs to be handled.
Hope this can be useful.

harshavardhana: This is a good idea @Yurmusp will test it out and see if there are any problems for 3.x

Yurmusp: Not sure what was meant. Who will test it?

harshavardhana: "will" means "we will" shorter way to saying it :-)

Yurmusp:
:)
OK. Sounds good.

Yurmusp:
Please, let me know when done.
How long you think it will take?

deekoder:
We have a process to review and test. As soon that's done it will be merged.

Yurmusp:
Is it days, weeks, months?
can you tell what is a ball park usually?

harshavardhana:
> Is it days, weeks, months?
> can you tell what is a ball park usually?

Usually 2-3 days.. @Yurmusp

Yurmusp:
Sounds good. Thank you.

deekoder:
@Yurmusp  its taking a bit longer because we are right now doing some fixes in minio/minio. Please stay tuned on this one.

Yurmusp:
Yes. Waiting.
Please, do not delay too much if possible.
Thanks.

Yurmusp:
Guys,

Any progress?

minio/api.py
Yurmusp: In my understanding. Change I suggested shall have no impact on 3.x. 
Because this change is actually to make everything more like 3.x 
(every string to be unicode) but within 2.x python.

harshavardhana: We won't be doing this for now this PR works as is.. We will take this activity as such an issue comes in future. We don't want to introduce subtle changes like this which can have unexpected consequences.

Yurmusp: As long as it works, it's fine I think.

Yurmusp:
When you guys expect it to be released?

Yurmusp:
Thank you guys. Looks like this works.

PR #531, opened 2017-06-11
Use hex instead of hash interface in partMetadata

vadmeste:
It would be little easier to use hex directly to express
a part's md5 or sha256 sum instead of hash interface. Convert
hex to base64 whenever we need that.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -37,6 +37,7 @@ import io
 import json
 import os
 import itertools
+import codecs
 
 # Dependencies
 import urllib3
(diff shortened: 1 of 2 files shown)


minio/api.py
```diff
@@ -1558,7 +1559,7 @@ def _do_put_multipart_object(self, bucket_name, object_name, part_metadata,
 
         headers = {
             'Content-Length': part_metadata.size,
-            'Content-Md5': part_metadata.md5hasher.base64digest()
+            'Content-Md5': base64.b64encode(part_metadata.md5_hex.decode("hex"))
```
poornas: For python3 , this does not work - I get an AttributeError: 'str' object has no attribute 'decode' part_metadata.md5_hex  is of type string.

vadmeste: > For python3 , this does not work - I get an AttributeError: 'str' object has no attribute 'decode' part_metadata.md5_hex is of type string.

Weird, I didn't get any error when I tested `python3 examples/fput_object.py`, can you tell me how did you test exactly ?

poornas: The error happens if you run python3 fput_object.py, i.e. by moving the fput_object.py from examples/ folder to the main repo directory.

harshavardhana: Basically this would import your current code @vadmeste

vadmeste: @poornas, the issue is fixed, can you check/review again ?

vadmeste:
@harshavardhana can you review this again and merge?

PR #536, opened 2017-06-15
multipart: Add support of parallel part uploads

### LINKED ISSUE #503: Parallel upload through fput_object API
Uploading a 7.5 gig file through mc client takes 30 seconds, while minio-py API fput_object takes about 3 minutes for the same task.
There should be an option for parallel upload in the API as well which can match the mc client performance.

vadmeste:
Fixes #503


harshavardhana:
If this is not a WIP can you update the PR title @vadmeste ?

vadmeste:
> If this is not a WIP can you update the PR title @vadmeste ?

Not WIP anymore, I had to test it.

minio/api.py
```diff
@@ -615,58 +674,28 @@ def fput_object(self, bucket_name, object_name, file_path,
...
-            part = LimitedReader(file_data, current_part_size)
-            part_metadata = PartMetadata(part, md5hasher.hexdigest(), sha256hasher.hexdigest(),
-                                         current_part_size)
-            # Initiate multipart put.
-            etag = self._do_put_multipart_object(bucket_name, object_name,
-                                                 part_metadata,
-                                                 upload_id,
-                                                 part_number)
+        # Run part uploads in parallel
+        part_uploads_result = p.map(self._upload_part, parts_to_upload)
```
donatello: Would'nt it be better to quit on first permanent error? As an example scenario, let's say that server goes down as soon as upload starts, each thread here will wait until it times out, but we know that if a particular upload has a permanent failure (all retries failed, etc), we can quickly exit the routine and report the failure. In the previous implementation when an exception is thrown the loop exits immediately. From the lib docs on multiprocessing, I am not sure what the error behaviour is.

minio/api.py
```diff
@@ -553,6 +559,59 @@ def _get_upload_id(self, bucket_name, object_name, metadata=None):
 
         return upload_id
 
+    def _upload_part(self, part_info):
+
+        uploaded_parts, file_path, bucket_name, object_name, upload_id, part_number, part_size = \
+                part_info[0], part_info[1], part_info[2], part_info[3], \
+                part_info[4], part_info[5], part_info[6]
```
poornas: wouldn't uploaded_parts, file_path, bucket_name, object_name, upload_id, part_number, part_size = part_info work as well?

minio/api.py
```diff
@@ -615,58 +674,28 @@ def fput_object(self, bucket_name, object_name, file_path,
             # Save uploaded parts for future verification.
             uploaded_parts[part.part_number] = part
 
-        # Always start with first part number.
+        # Threads pool which upload parts
+        p = multiprocessing.dummy.Pool(_PARALLEL_UPLOADERS)
```
poornas: pool rather than p would be more descriptive.

minio/api.py
vadmeste: > Would'nt it be better to quit on first permanent error? As an example scenario, let's say that server goes down as soon as upload starts, each thread here will wait until it times out, but we know that if a particular upload has a permanent failure (all retries failed, etc), we can quickly exit the routine and report the failure. In the previous implementation when an exception is thrown the loop exits immediately. From the lib docs on multiprocessing, I am not sure what the error behaviour is.

You are right @donatello, also I noticed that if a thread crashes, it will crash all the program.
Even catching exception of fput_object API won't work.

```
try:
    client.fput_object('vadmeste', 'my-objectname', 'my-filepath')
except ResponseError as err:
    print(err)
```

 I'll try to find a workaround.

vadmeste: @donatello, it is possible to catch the exception from the code calling our fput_object. I made a typo and I didn't saw it. Multiprocessing stops sending work to the threads pool when an uncaught exception is raised. 

However, I noticed another problem which is that multiprocessor in python 2 doesn't detect keyboard interruption (bug) but python 3 does. I might have to change everything due to this issue..

vadmeste:
@donatello, I added thread_pool.py which can detect keyboard interruption and can also quit fast when one part upload encounters a problem such as, server or connection down.

@poornas We don't need a new functional test for this change since fput_object multipart is already tested.

I made several changes, please take time testing this.

harshavardhana:
> @donatello, I added thread_pool.py which can detect keyboard interruption and can also quit fast when one part upload encounters a problem such as, server or connection down.
> 
> @poornas We don't need a new functional test for this change since fput_object multipart is already tested.
> 
> I made several changes, please take time testing this.

Can you also remove the resuming @vadmeste ?

minio/thread_pool.py
```diff
@@ -0,0 +1,80 @@
...
+        self.results_queue = queue()
+        self.tasks_queue = queue(num_threads)
+        for _ in range(num_threads):
+            Worker(self.tasks_queue, self.results_queue)
+
+    def add_task(self, func, *args, **kargs):
+        """ Add a task to the queue """
+        self.tasks_queue.put((func, args, kargs))
+
+    def map(self, func, args_list):
```
donatello: It is not correct to call this implementation `map` - in python (and in a bunch of other languages) the meaning of map is calling a function over an array of arguments and collecting the results. The order of the results has to correspond to the order of the arguments. This is not the case in this implementation.

You could call the function something else, or perhaps re-implement the function.

minio/thread_pool.py
```diff
@@ -0,0 +1,80 @@
...
+
+class ThreadPool:
+    """ Pool of threads consuming tasks from a queue """
+    def __init__(self, num_threads):
+        self.results_queue = queue()
+        self.tasks_queue = queue(num_threads)
+        for _ in range(num_threads):
+            Worker(self.tasks_queue, self.results_queue)
+
+    def add_task(self, func, *args, **kargs):
```
donatello: Any reason ThreadPool needs to implement these particular functions? It looks like this function is just a one-liner and could be eliminated?

vadmeste:
I rewrote some stuffs in this PR, can you review again @poornas  and @donatello  ?

harshavardhana:
what is pending on this one @vadmeste ?

vadmeste:
> what is pending on this one @vadmeste ?

PR rebased. Waiting for @donatello @poornas inputs

harshavardhana:
https://github.com/minio/minio-py/pull/546 once this PR is merged build issues will go away.

vadmeste:
@donatello @poornas, I made only few changes since last PR, can you review again?

minio/thread_pool.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def run(self):
+        fast_quit = False
+        while True:
+            func, args, kargs = self.tasks_queue.get()
+            if not fast_quit:
+                try:
+                    result = func(*args, **kargs)
+                    self.results_queue.put(result)
+                except Exception as e:
```
donatello: This handler is catching all exceptions and ignoring them.

harshavardhana:
This is currently blocked on @vadmeste evaluating asyncio support in python3

harshavardhana:
Were you able to validate what we last spoke @vadmeste ? if you are not convinced with asyncio can you rebase this PR? .. We can perhaps move forward on merging this.

vadmeste:
> Were you able to validate what we last spoke @vadmeste ? if you are not convinced with asyncio can you rebase this PR? .. We can perhaps move forward on merging this.

https://github.com/minio/minio-py/compare/master...vadmeste:issues/st-503?expand=1

This is the branch that already works.. you can review it (especially thread_pool_py3.py).. 

asyncio needs "native" coding using "async" and "await" keywords. The developer explicitly says run this function asynchronously or wait for the result of this async function, if we use these keywords, we will need many code for api.py to be duplicated for py2 and py3.

However, my work avoids this (see thread_pool_py2 and thread_pool_py3).. if we agree that the code is easy to read, I'll rebase this PR and make it ready for merge

poornas:
@vadmeste , needs rebase

harshavardhana:
> https://github.com/minio/minio-py/compare/master...vadmeste:issues/st-503?expand=1
> 
> This is the branch that already works.. you can review it (especially thread_pool_py3.py)..

I think we can stick to what is being done here.. unless @donatello has any arguments.

vadmeste:
> https://github.com/minio/minio-py/compare/master...vadmeste:issues/st-503?expand=1
> 
> This is the branch that already works.. you can review it (especially thread_pool_py3.py)..
> I think we can stick to what is being done in this PR avoiding asyncio.. unless @donatello has any arguments.

@donatello, can you comment this ? ^^

donatello:
@vadmeste I think the async io code looks good and works but needs cleanup. In the meanwhile I was doing some experimentation and found a more simpler _potential_ solution using thread-pool library functionality: https://gist.github.com/donatello/9b6954d74cd60356507bd5ea44eaa396 - main part:

```py
_is_py3 = (sys.version_info[0] == 3)

if _is_py3:
    from concurrent.futures import ThreadPoolExecutor
    tp = ThreadPoolExecutor(max_workers=_PARALLEL_UPLOADERS)
else:
    # Python 2 :(
    from multiprocessing.dummy import Pool
    tp = Pool(_PARALLEL_UPLOADERS)

```

The gist has a job function that throws exceptions - it is caught properly in the `try` block around the map function. The exception handling is much easier with code like this. Please try running the code in the gist in both py2 and py3.

donatello:
On further investigation it turns out that `concurrent.futures.ThreadPoolExecutor` is a threading library based module and does not internally use `asyncio`.

Also, I learned a few things about asyncio :

1. Using async calls does not make network/disk calls asynchronous automatically - using a library that does these things asynchronously from the ground-up is required. This means network/disk inside the event loop actually block the event loop unless we are using libraries like [aiohttp](https://aiohttp.readthedocs.io/en/stable/) and aiofiles (appears to be a newish library).
2. Due to the above, use of `concurrent.futures.ThreadPoolExecutor` is required to fire-off such blocking calls in a separate thread to make them _appear_ asynchronous, and to integrate nicely with asyncio. This means that there will be thread-switching inside the executor - this was something we wanted to avoid in the first place.

Due to these problems it would not be a clean solution for py3. It would be better to reconsider this option in the future and see if a cleaner solution appears.

An interesting point is that a high-quality boto based async s3 library is [here](https://github.com/aio-libs/aiobotocore).

So, IMO we should go ahead with merging this PR.

The exception handling in the current PR is not correct - the exception inside a task needs to be rethrown in the main thread, instead of being just printed and otherwise ignored. It appears that `multiprocessing.dummy.Pool` would work in both py3 and py2 and it re-raises exceptions from threads in the calling thread. If there are no other issues, we should go back to that simple solution.

vadmeste:
> Using async calls does not make network/disk calls asynchronous automatically

Why ? Using async + await is already able to run some code in parallel without using threads.



> Due to the above, use of concurrent.futures.ThreadPoolExecutor is required to fire-off such blocking calls in a separate thread to make them appear asynchronous, and to integrate nicely with asyncio. This means that there will be thread-switching inside the executor - this was something we wanted to avoid in the first place.

Yes you are right, my approach on this is to use only one additional thread instead of a new thread for each worker.

> The exception handling in the current PR is not correct

oups, yes. I'll fix it.

harshavardhana:
why closed? @vadmeste

PR #540, opened 2017-06-20
tests: put_object a big file to cover more code

### LINKED ISSUE #539: Add one test which covers _stream_put_object()
We made a mistake in one change in _stream_put_object() and functional tests wasn't able to detect it. So we need to add a test which covers  _stream_put_object().

vadmeste:
Fixes #539

### DIFF
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 # Minio Python Library for Amazon S3 Compatible Cloud Storage,
-# (C) 2015, 2016 Minio, Inc.
+# (C) 2015, 2016, 2017 Minio, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -20,6 +20,10 @@ import io
 import uuid
 import urllib3
 import certifi
+import random
+
+from random import choice
+from string import ascii_lowercase
 
 from datetime import datetime, timedelta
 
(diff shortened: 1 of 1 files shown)


tests/functional/tests.py
```diff
@@ -85,21 +86,29 @@ def main():
...
-
+    # Create a test file
     with open(testfile, 'wb') as file_data:
         file_data.write(fake.text().encode('utf-8'))
     file_data.close()
 
+    # Create a large file
+    with open(largefile, 'wb') as file_data:
+        for i in range(0, 104857):
+            file_data.write(fake.text().encode('utf-8'))
```
harshavardhana: Just use zero bytes instead of using `fake.text()` would be faster.. and avoids CPU computation.

tests/functional/tests.py
```diff
@@ -85,21 +86,29 @@ def main():
     newfile_f = 'newfile-f 新'
     newfile_f_custom = 'newfile-f-custom'
 
-
+    # Create a test file
     with open(testfile, 'wb') as file_data:
         file_data.write(fake.text().encode('utf-8'))
     file_data.close()
```
krisis: `with` statement closes the file, we don't need to close it explicitly.

tests/functional/tests.py
```diff
@@ -85,21 +86,29 @@ def main():
...
 
+    # Create a large file
+    with open(largefile, 'wb') as file_data:
+        for i in range(0, 104857):
+            file_data.write(fake.text().encode('utf-8'))
+    file_data.close()
```
krisis: Same as above, we don't need to close the file, `with` statement closes it.

vadmeste: > Same as above, we don't need to close the file, with statement closes it.

Fixed

tests/functional/tests.py
vadmeste: > Just use zero bytes instead of using fake.text() would be faster.. and avoids CPU computation.

We should avoid zero bytes because in that case we won't check if parts are concatenated with the correct order. I made changes in the code (see random_data class) to generate pseudo random data much more faster.

vadmeste:
I made more changes to this PR @krisis & @harshavardhana, can you review again ?

PR #542, opened 2017-06-21
multipart: Disable resuming incomplete uploads

### LINKED ISSUE #541: Disable resuming uploads
Do not fetch old uploads id and resume them. Instead create a new upload id each time we upload a new multipart.

vadmeste:
Fixes #541

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -525,35 +525,6 @@ class Minio(object):
                 response.close()
                 continue
 
-    def _get_upload_id(self, bucket_name, object_name, metadata=None):
-        """
-        Get previously uploaded upload id for object name or initiate a request
-        to fetch a new upload id.
-
-        :param bucket_name: Bucket name where the incomplete upload resides.
-        :param object_name: Object name for which upload id is requested.
-        :param metadata: Additional metadata headers for new multipart upload.
-        """
-        recursive = True
-        current_uploads = self._list_incomplete_uploads(bucket_name,
-                                                        object_name,
-                                                        recursive,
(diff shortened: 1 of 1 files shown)


vadmeste:
Now, fput_object just calls put_object. I don't see why not, also tests are passing.

 @harshavardhana, can you comment on this ? I will made more tests if you don't contest this change.

vadmeste:
This change is not WIP anymore. fput_object and put_object have common code path now.

fput_object() a f1gb file consumes RSS 32MB in contrary to RSS 16MB before this PR.

I think we can accept this compromise to avoid code duplication.

donatello:
> fput_object() a f1gb file consumes RSS 32MB in contrary to RSS 16MB before this PR.
> I think we can accept this compromise to avoid code duplication.

I am ok with the compromise, but I would like to understand why the memory consumption went up. Could you explain @vadmeste ?

vadmeste:
> I am ok with the compromise, but I would like to understand why the memory consumption went up. Could you explain @vadmeste ?

@donatello, put_object considers that the input reader is not seekable, so it allocates memory to save data to calculate hash and then send data to the server via PUT part request.

fput_object plays with file.seek(), it reads the file twice (again for hash and for PUT part) without having the obligation to save any data in the memory (syscalls and the kernel would take care of that).

Now, you would say that we save the whole data twice  in the memory.. that's true.. but it looks like the garbage collector frees data that we don't need anymore, that's what I concluded when I followed memory consumption during upload of 1Gb file.

PR #546, opened 2017-07-01
Simplify python tests and reduce CPU usage.

harshavardhana:


vadmeste:
@harshavardhana, you probably need to remove reduce CPU usage in commit message

harshavardhana:
No @vadmeste there is a reason why it's there .  Check locally with this change and without . Look at total time taken for completion of tests.

deekoder:
Can this be merge?

harshavardhana:
> Can this be merged?

Yes @deekoder

PR #547, opened 2017-07-04
tests: Add support for functional tests to run in mint.

harshavardhana:
This PR is needed to support multiple environments in
which the following test script will be run.


harshavardhana:
This is ready for review @krisis @NitishT @vadmeste with mint changes.

tests/functional/tests.py
```diff
@@ -24,29 +24,46 @@
...
+
+        :param amt:
+            How much of the content to read.
+        """
+        # If offset is bigger than size. Treat it as EOF return here.
+        if self._offset_location + amt > self._limit:
+            # return empty bytes to indicate EOF.
+            return b''
+
+        data = b''.join([b'o' for i in range(amt)])
```
krisis: this doesn't look like random data.

tests/functional/tests.py
```diff
@@ -102,108 +119,101 @@ def main():
...
+        largefile = '/mint/data/datafile-65-MB'
+        # fput a file
+        print("Upload a local file")
+        client.fput_object(bucket_name, object_name+'-f', testfile)
+        if is_s3:
+            client.fput_object(bucket_name, object_name+'-f', testfile,
+                               metadata={'x-amz-storage-class': 'STANDARD_IA'})
+
+        # Fput a large file.
+        print("Upload a largfile through multipart upload")
```
krisis: typo: s/largfile/largefile

tests/functional/tests.py
harshavardhana: My bad..local change is not pushed.

tests/functional/tests.py
harshavardhana: Done

NitishT:
@harshavardhana does this also address https://github.com/minio/minio-py/issues/548 ?

harshavardhana:
> @harshavardhana does this also address #548 ?

Yes it does @NitishT

harshavardhana:
Changed a bunch of things, added more randomness as asked by @krisis . Please take a look again.

tests/functional/tests.py
```diff
@@ -16,37 +16,56 @@
...
+            How much of the content to read.
+        """
+        # If offset is bigger than size. Treat it as EOF return here.
+        if self._offset_location + amt > self._limit:
+            # return empty bytes to indicate EOF.
+            return b''
+
+        # make translation table from 0..255 to 97..122
+        bal = [c.encode('ascii') for c in ascii_lowercase]
+        data = b''.join([bal[int(random() * 26)] for _ in range(amt)])
```
krisis: As a consequence to the previous comment, we should compute the remaining bytes to be returned.
`amt = min(amt, self._limit - self._offset_location)`

tests/functional/tests.py
```diff
@@ -16,37 +16,56 @@
...
+
+    def read(self, amt=64*1024):
+        """
+        Similar to :meth:`io.read`, with amt option.
+
+        :param amt:
+            How much of the content to read.
+        """
+        # If offset is bigger than size. Treat it as EOF return here.
+        if self._offset_location + amt > self._limit:
```
krisis: we should return empty bytes (`b''`) only when `self._offset_location == self._limit`. E.g, when
`self._limit = 10`, `self._offset = 9` and `amt = 5`, we should return the last byte and not `b''`.

harshavardhana:
For path APIs we need local files. Streaming is only for generating content for streaming APIs

krisis:
@harshavardhana IIUC, we are running file based object operations (like FPutObject) inside mint environment and stream-based object operations outside of mint environment. Shouldn't we run all kinds of operations in both environments?

harshavardhana:
> @harshavardhana IIUC, we are running file based object operations (like FPutObject) inside mint environment and stream-based object operations outside of mint environment. Shouldn't we run all kinds of operations in both environments?

Yes ideally, but there seems to be some confusing aspects floating around on the nature of functional tests themselves.  Should be used to test SDK or server and what is their rightful place.

Updated the PR as well to provide mint as a sub-set behavior only meant to be used as data dir location. If not we auto-generate this data such that some APIs are not skipped on regular testing without mint.

PR #550, opened 2017-07-06
api: put_object shouldn't calculate md5sum always.

### LINKED ISSUE #511: For all signed payloads with sha256 avoid calculating Content-Md5
Don't compute md5sum signatures when uploading for the first time, only compute when resuming.

harshavardhana:
- content-md5 is calculated and set if HTTPS server
  endpoint is used.
- content_sha256 is calculated and set if HTTP server
  endpoint is used.

Fixes #511

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -65,14 +65,15 @@ from .helpers import (get_target_url, is_non_empty_string,
                       is_valid_endpoint,
                       get_sha256_hexdigest, get_md5_base64digest, Hasher,
                       optimal_part_info,
-                      is_valid_bucket_name, PartMetadata, parts_manager,
+                      is_valid_bucket_name, PartMetadata, read_full,
                       is_valid_bucket_notification_config,
                       mkdir_p, dump_http)
 from .helpers import (MAX_MULTIPART_OBJECT_SIZE,
                       MIN_OBJECT_SIZE)
 from .signer import (sign_v4, presign_v4,
                      generate_credential_string,
-                     post_presign_signature, _SIGN_V4_ALGORITHM)
+                     post_presign_signature)
+from .signer import (_UNSIGNED_PAYLOAD, _SIGN_V4_ALGORITHM)
 from .xml_marshal import (xml_marshal_bucket_constraint,
                           xml_marshal_complete_multipart_upload,
                           xml_marshal_bucket_notifications,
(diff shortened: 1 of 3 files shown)


harshavardhana:
Looks like there is a bug in the change here.. will fix it and perhaps then we can start review.

harshavardhana:
> Looks like there is a bug in the change here.. will fix it and perhaps then we can start review.

This is fixed - reviews welcome!

minio/helpers.py
```diff
@@ -137,20 +135,16 @@ def parts_manager(data, part_size=5*1024*1024):
     :param part_size: Individual part number defaults to 5MB.
     :return: Returns :class:`PartMetadata <PartMetadata>`
     """
-    tmpdata = io.BytesIO()
-    md5hasher = Hasher.md5()
-    sha256hasher = Hasher.sha256()
-    total_read = 0
-    while total_read < part_size:
+    part_data = io.BytesIO()
```
vadmeste: Can you check if part_size % 1024 == 0 and throw an error otherwise ? because we are reading with 1024 bytes chunk and this algorithm won't be correct if we change that read chunk size

minio/api.py
```diff
@@ -1400,8 +1404,9 @@ def _presigned_get_partial_object(self, bucket_name, object_name,
                                  expires=int(expires.total_seconds()))
         return presign_url
 
-    def _do_put_multipart_object(self, bucket_name, object_name, part_metadata,
-                                 upload_id='', part_number=0, metadata=None):
+    def _do_put_object(self, bucket_name, object_name, part_data,
+                       part_size, upload_id='', part_number=0,
+                       metadata=None):
         """
         Initiate a multipart PUT operation for a part number.
```
vadmeste: Can you update this block? as I understood this function can upload a single put or a part

minio/helpers.py
harshavardhana: Why is that ? 1024 is just to fill in 5MB

minio/api.py
harshavardhana: Will do.

minio/helpers.py
vadmeste: > 1024 is just to fill in 5MB

yes, but if a developer modifies this line:
 
`current_data = data.read(1024)`  
to 
`current_data = data.read(1024*3)`

the function won't work anymore like expected.

harshavardhana: i don't think so.. i was able to run it fine. @vadmeste

vadmeste: > i don't think so.. i was able to run it fine. @vadmeste

It is weird.. we can't fill 5mb with 1024*3 sized chunk since (5*1024*1024) % (1024*3) != 0

harshavardhana:
@balamurugana you might need to approve again.

PR #555, opened 2017-07-12
Split tests into functions

### LINKED ISSUE #551: Update test cases for minio-py sdk
Mint pulls functional test cases from minio-py sdk.

Minio-Py SDK tests should be updated to address the following

- Named and commented to indicate the use case being tested.
- Access all the public Minio server APIs via minio-py.
- Include one positive case and couple of boundary cases for each API.
- Use streaming files as preferred way to generate files for testing. (This can be a ref: https://github.com/minio/minio-java/pull/586) 
- Use static files only in case where streaming can't be used. 

This is related to https://github.com/minio/mint/issues/43

vadmeste:
Fixes #551

### DIFF
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -15,6 +15,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
+sys.path.append('/home/vadmeste/work/python/minio-py')
+
 import os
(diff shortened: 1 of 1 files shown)


vadmeste:
New changes are introduced to satisfy both @harshavardhana and @krisis comments.

PR #559, opened 2017-07-21
Fix get_bucket_policy crash for TypeError

### LINKED ISSUE #558: get_policy call fails with TypeError
When functional tests are run against azure gateway, get_policy call fails with a TypeError. The client sdk does not guard against server running None for statements while appending policy statements in policy.py

File "/usr/bin/minio-py-functional", line 333, in <module>
main()
File "/usr/bin/minio-py-functional", line 277, in main
policy_name = client.get_bucket_policy(bucket_name)
File "/usr/local/lib/python3.5/dist-packages/minio/api.py", line 368, in get_bucket_policy
policy._append_statements(statements, policy_dict.get('Statement', []))
File "/usr/local/lib/python3.5/dist-packages/minio/policy.py", line 429, in _append_statements
for s in append_statements:
TypeError: 'NoneType' object is not iterable

poornas:
Fixes #558 
get_bucket_policy  function crashes with TypeError when azure gateway returns null statements in Policy
{"Version":"2012-10-17","Statement":null}. Check value returned before appending to statements list.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -365,6 +365,8 @@ class Minio(object):
         if not policy_dict:
             return policy.Policy.NONE
 
+        if policy_dict.get('Statement') is None:
+            raise ValueError("None Policy statement")
         # Normalize statements.
         statements = []
         policy._append_statements(statements, policy_dict.get('Statement', []))


donatello:
I think this PR is not needed - https://github.com/minio/minio-py/issues/558#issuecomment-317138349 - I believe the issue is with the server.

poornas:
@donatello and @balamurugana , updated PR to have SDK handle empty statements from server  by returning None as PolicyType rather than crash.

minio/api.py
```diff
@@ -365,6 +365,9 @@ def get_bucket_policy(self, bucket_name, prefix=""):
         if not policy_dict:
```
balamurugana: You could fix it like
```python
if not policy_dict or not policy_dict.get('Statement'):
    return policy.Policy.NONE
```

minio/api.py
```diff
@@ -383,7 +386,9 @@ def set_bucket_policy(self, bucket_name, prefix, policy_access):
         policy_dict = self._get_bucket_policy(bucket_name)
         if policy_access == policy.Policy.NONE and not policy_dict:
             return
-
+        if policy_dict.get('Statement', []) is None:
+            return
```
balamurugana: for 'set_bucket_policy()` you shouldn't return without setting it.

minio/api.py
```diff
@@ -383,7 +386,9 @@ def set_bucket_policy(self, bucket_name, prefix, policy_access):
...
+        
         if not policy_dict:
             policy_dict = {'Statement': [],
                            "Version": "2012-10-17"}
```
balamurugana: the fix could be added here like
```python
if not policy_dict:
    policy_dict = {'Statement': [],
                   "Version": "2012-10-17"}
elif not policy_dict.get('Statement'):
    policy_dict['Statement'] = []
```

minio/api.py
```diff
@@ -383,11 +383,12 @@ def set_bucket_policy(self, bucket_name, prefix, policy_access):
         policy_dict = self._get_bucket_policy(bucket_name)
         if policy_access == policy.Policy.NONE and not policy_dict:
             return
-
+
```
balamurugana: You could fix this lint error.

minio/api.py
```diff
@@ -383,11 +383,12 @@ def set_bucket_policy(self, bucket_name, prefix, policy_access):
...
         if policy_access == policy.Policy.NONE and not policy_dict:
             return
-
+        
         if not policy_dict:
             policy_dict = {'Statement': [],
                            "Version": "2012-10-17"}
-
+        elif not policy_dict.get('Statement'):
+            policy_dict['Statement'] = []
```
balamurugana: You could add a blank line after this `if` block.

donatello:
I still think this PR is not doing the right thing - if the server sends an invalid policy (one with an empty policy statement), we do not want to crash - this is understandable. But we should be throwing a Python `ValueError` and not converting it into a `NONE` policy. The definition of a `NONE` policy is when the server returns 404 http error code for a get-bucket-policy API call. If we convert something invalid into a valid value like this, we are hiding a bug that ought to be fixed.

poornas:
@donatello, done.

poornas:
@donatello, the functional tests should still not mask this error  by catching the exception and passing the test.Have raised a PR for fixing Azure response.

harshavardhana:
> @donatello, the functional tests should still not mask this error by catching the exception and passing the test.Have raised a PR for fixing Azure response.

But the tests are running against play.minio.io so it must be that there is a bug in the FS mode as well? @poornas

poornas:
@donatello, fixed.

minio/api.py
```diff
@@ -365,6 +365,8 @@ def get_bucket_policy(self, bucket_name, prefix=""):
         if not policy_dict:
             return policy.Policy.NONE
 
+        if not policy_dict.get('Statement'):
```
balamurugana: This is not the right fix.

```py
>>> policy_dict = {'Statement': []}
>>> print not policy_dict.get('Statement')
True
>>> policy_dict = {'Statement': None}
>>> print not policy_dict.get('Statement')
True
```

We don't need to throw error on empty statement list.

minio/api.py
```diff
@@ -365,6 +365,8 @@ def get_bucket_policy(self, bucket_name, prefix=""):
...
+            raise ValueError("Policy statement cannot be empty")
```
balamurugana: This error message is not right too.  As we received invalid JSON from the server, None value set to policy_dict['Statement'].  You would need to indicate the error something like `None policy statement`

poornas:
@balamurugana ,done

PR #563, opened 2017-08-05
Parallel upload of object parts

### LINKED ISSUE #503: Parallel upload through fput_object API

vadmeste:
Fixes #503 

Based on the first attempt here https://github.com/minio/minio-py/pull/536

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -69,7 +69,7 @@ from .helpers import (get_target_url, is_non_empty_string,
                       is_valid_bucket_notification_config,
                       mkdir_p, dump_http)
 from .helpers import (MAX_MULTIPART_OBJECT_SIZE,
-                      MIN_OBJECT_SIZE)
+                      MIN_PART_SIZE)
 from .signer import (sign_v4, presign_v4,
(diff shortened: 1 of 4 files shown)


tests/functional/tests.py
```diff
@@ -15,9 +15,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import sys
-sys.path.append('/home/vadmeste/work/python/minio-py')
```
harshavardhana: oh wow when was this added?

vadmeste: > oh wow when was this added?

Few weeks ago, it was when I restructured tests, which looks like a big change in diff.

harshavardhana: Can you send fix for this quickly? lets get it merged before merging this PR?

vadmeste:
@donatello, can you review this ?

PR #564, opened 2017-08-06
tests: Remove useless path append

vadmeste:
Path append was accidentally merged in the code, it was only helping me
to test locally.


vadmeste:
Can this PR be merged?

PR #581, opened 2017-09-26
Do not fail input checks for endpoints with .amazonaws.com as prefix

### LINKED ISSUE #570: minio-py fails erroneously with minio.error.InvalidEndpointError: InvalidEndpointError: message: Amazon S3 hostname should be s3.amazonaws.com
I have minio installed to our kubernetes cluster on amazon. The hostname ends in amazonaws.com but is not S3. When I run the minio-py with the correct credentials, the code fails here:

https://github.com/minio/minio-py/blob/master/minio/helpers.py#L276-L279

The check for s3 ignores the fact that I may have an external IP exposed through amazon that is not s3, but rather minio - in this case hosted on our kubernetes cluster.

Do you need a pull request?

harshavardhana:
Additionally also support extracting region from the endpoint, this
technique is an optimization avoids GetBucketLocation() request
and makes it more performant.

Fixes #570

### DIFF
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # Minio Python Library for Amazon S3 Compatible Cloud Storage, (C)
-# 2015, 2016 Minio, Inc.
+# 2015, 2016, 2017 Minio, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -19,7 +19,7 @@ minio.helpers
 
 This module implements all helper functions.
 
-:copyright: (c) 2015 by Minio, Inc.
+:copyright: (c) 2015, 2016, 2017 by Minio, Inc.
 :license: Apache 2.0, see LICENSE for more details.
 
 """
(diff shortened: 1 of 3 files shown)


minio/helpers.py
```diff
@@ -49,7 +49,24 @@
 _ALLOWED_HOSTNAME_REGEX = re.compile(
     '^((?!-)[A-Z\\d-]{1,63}(?<!-)\\.)*((?!-)[A-Z\\d-]{1,63}(?<!-))$',
     re.IGNORECASE)
+_EXTRACT_REGION_REGEX = re.compile('s3-(.+?).amazonaws.com')
```
poornas: regex wont work if endpoint is like s3.us-west-1.amazonaws.com

harshavardhana:
@donatello need your reviews.

harshavardhana:
@donatello pending your review.

minio/helpers.py
```diff
@@ -49,7 +49,24 @@
 _ALLOWED_HOSTNAME_REGEX = re.compile(
     '^((?!-)[A-Z\\d-]{1,63}(?<!-)\\.)*((?!-)[A-Z\\d-]{1,63}(?<!-))$',
     re.IGNORECASE)
+_EXTRACT_REGION_REGEX = re.compile('s3[.-]?(.+?).amazonaws.com')
```
donatello: There is one more variant that seems easy to add here e.g. `s3.dualstack.us-east-2.amazonaws.com` - see https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region

harshavardhana: done

harshavardhana:
@poornas need your review again.

PR #582, opened 2017-09-28
tests: Return NA when APINotImplemented for bucket policy

### LINKED ISSUE #579: get_bucket_policy functional test fails for azure and gcs
Mint fails for minio-py sdk tests on get_bucket_policy for Azure and GCS. These tests should be skipped conditionally when run for Azure /GCS and the output log reflect that the test has been skipped until policy support is implemented.
```
  File "/mint/run/core/minio-py/tests.py", line 333, in <module>
    main()
  File "/mint/run/core/minio-py/tests.py", line 277, in main
    policy_name = client.get_bucket_policy(bucket_name)
  File "/usr/local/lib/python3.5/dist-packages/minio/api.py", line 368, in get_bucket_policy
    policy._append_statements(statements, policy_dict.get('Statement', []))
  File "/usr/local/lib/python3.5/dist-packages/minio/policy.py", line 429, in _append_statements
    for s in append_statements:
TypeError: 'NoneType' object is not iterable
```

Mint image built from  commit 158108152fc86c52d5543995e0e7634bf1b6da74
Minio version:
Version: 2017-09-25T21:57:17Z
Release-Tag: RELEASE.2017-09-25T21-57-17Z
Commit-ID: d3eb5815d9951fdb82d0134b7651368bab71d72e

vadmeste:
We need this temporary behavior is Mint can succeed. This will be revert
once bucket policy will be implemented in all gateways.

Fixes #579

### DIFF
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -32,7 +32,7 @@ import certifi
 
 from minio import Minio, PostPolicy, CopyConditions
 from minio.policy import Policy
-from minio.error import (ResponseError, PreconditionFailed,
+from minio.error import (APINotImplemented, ResponseError, PreconditionFailed,
                          BucketAlreadyOwnedByYou, BucketAlreadyExists)
 
 class LimitedRandomReader(object):
@@ -98,10 +98,10 @@ class LogOutput(object):
             'status': exit status, possible values are 'PASS', 'FAIL', 'NA',
                       defaults to 'PASS'
     """
-class LogOutput(object):
 
     PASS = 'PASS'
     FAIL = 'FAIL'
+    NA = 'NA'
 
     def __init__(self, meth):
         self.__args_list = inspect.getargspec(meth).args[1:]
@@ -128,7 +128,7 @@ class LogOutput(object):
     @args.setter
     def args(self, val): self.__args = val
 
-    def json_report(self, err_msg=''):
+    def json_report(self, err_msg='', status=''):
         self.__args = {k: v for k, v in self.__args.items() if v and v != ''}
         entry = {'name': self.__name,
             'function': self.__function,
@@ -137,7 +137,8 @@ class LogOutput(object):
             'alert': self.__alert,
             'message': str(err_msg),
             'error': traceback.format_exc() if err_msg and err_msg != '' else '',
-            'status': self.FAIL if err_msg and err_msg != '' else self.PASS
+            'status': status if status and status != '' else \
+                    self.FAIL if err_msg and err_msg != '' else self.PASS
         }
         return json.dumps({k: v for k, v in entry.items() if v and v != ''})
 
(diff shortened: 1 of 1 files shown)


tests/functional/tests.py
```diff
@@ -602,6 +602,8 @@ def test_get_bucket_policy(client, log_output):
         policy_name = client.get_bucket_policy(bucket_name)
         if policy_name != Policy.NONE:
             raise ValueError('Policy name is invalid: ' + policy_name)
+    except APINotImplemented:
+        pass
```
ebozduman: Mint logging requirement defined a 3rd value for the `status` field, `NA` or not applicable, other than `PASS` and `FAIL`. I think `NA` can be used with a message `API is not implemented yet` in the`alert` field for this case. 

At least this way, the tester will be notified about the condition and see the reason why test did not pass or fail, instead of being completely silent about the test.
So, I suggest to have a `print(log_output.json_report_not_applicable())` method defined and called in the main except area to handle `APINotImplemented` exception. This could be done blindly for all tests without any need to know what APIs have been implemented or not.

vadmeste: It is better that we define first what's NA.

For me, it means, the test function decides to avoid to run tests since it doesn't make sense to do that in the current context.

In our case, we return NA because it doesn't make sense to run bucket policy get/set tests because we are in gateway mode.

But for that, we need to pass a parameter indicating that the server is a in gateway mode. Always interpreting an API NOT IMPLEMENTED as a NA is not correct in my point of view.

balamurugana: 1. `NA` is followed currently for Ignored cases due to `NotImplemented` S3 error.
2. As per our discussion we decided to ignore `NotImplemented` error for any S3 call irrespective of the server.  This means `mint` or functional tests of SDKs/tools  do not expect a parameter to control `NotImplemented`.  In a practical case, ignoring get and set bucket policy tests are good enough.
3. No test is written/ignored specifically for gateway.  If we so, this will lead to maintain states inside tests which will not be clean.  In future S3 gateway could have full support of policy where Azure/GCS gateway may support to certain degree.

vadmeste: so @balamurugana, do you recommend any change in this PR ?

vadmeste:
PR updated, please review again @ebozduman and @nitisht

vadmeste:
ping  @ebozduman and @nitisht

deekoder:
Please rebase @vadmeste

vadmeste:
@deekoder, I had to write this PR since other changes affecting this PR were merged first.

@ebozduman, @nitisht: can you review again ?

tests/functional/tests.py
```diff
@@ -102,6 +102,7 @@ class LogOutput(object):
```
ebozduman: This comment is actually for the above line#101 and I think it introduced this line by mistake.
`class LogOutput(object):` is already at line#75 and hence this one needs to be removed.

vadmeste: > This comment is actually for the above line#101 and I think it introduced this line by mistake.
> class LogOutput(object): is already at line#75 and hence this one needs to be removed.

yes @ebozduman, it is an old problem, I've just fixed now

Issue #583, opened 2017-09-29
Fix minor bugs in logging format


kannappanr:
Make the json logs generated by minio-py more compliant with mint logging format

1. Convert the duration field to display in milliseconds
2. Remove the description field
3. Remove entries that contain null in json.

kannappanr:
code that fixes this issue is part of PR #584 . Hence closing the issue.

PR #584, opened 2017-09-29
Fix minor bugs in logging format

### LINKED ISSUE #576: Bucket name used in functional test should have a prefix
Bucket names used in functional test should have a prefix. This is required by mint to cleanup left over buckets. Currently minio-py uses a uuid as the bucket name. It needs to add a prefix.

kannappanr:
1. Add prefix to the bucket name
2. Convert the duration to milliseconds
3. Remove description field
4. Remove entries that contain null in json

Fixes #576, #583

### DIFF
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -84,8 +84,6 @@ class LogOutput(object):
                         put together with the method name:
                         <method>.__name__+'('+', '.join(args_list)+')'
                         e.g. 'remove_object(bucket_name, object_name)'
-            'description': short description for the test, optional,
-                           defaults to empty string, ''
             'args': method/api arguments with their values, in
                     dictionary form: {'arg1': val1, 'arg2': val2, ...}
             'duration': duration of the whole test in milliseconds,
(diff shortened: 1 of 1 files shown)


tests/functional/tests.py
```diff
@@ -122,45 +119,44 @@ def name(self): return self.__name
...
+            'duration': int(round((time.time() - self.__start_time)*1000)),
             'alert': self.__alert,
             'message': str(err_msg),
             'error': traceback.format_exc(),
             'status': self.FAIL
-        })
+        }
+        return json.dumps({k: v for k, v in entry.items() if v and v != ''})
 
     def json_report_success(self):
```
balamurugana: `json_report_failure()` and ` json_report_success()` are same code expect `err_msg`. You could use lower level function which accepts `err_msg` where `json_report_failure()` and `json_report_success()` uses the function by passing `err_msg` with null or value.

kannappanr:
@balamurugana Have simplified the function to use one function to log both success and failure

tests/functional/tests.py
```diff
@@ -122,45 +119,34 @@ def name(self): return self.__name
...
     def function(self, val): self.__function = val
-    @description.setter
-    def description(self, val): self.__description = val
     @args.setter
     def args(self, val): self.__args = val
 
-    def json_report_failure(self, err_msg):
-        # print(err_msg)
-        return json.dumps({'name': self.__name,
+    def json_report(self, err_msg = ''):
```
balamurugana: `err_msg=''` is pep8 compatible.

PR #586, opened 2017-10-10
Fix: Use dict.items() instead of dict.iteritems()

### LINKED ISSUE #585: Python3: multible policies wont work
Because policy.py uses iteritems() in line 381 and 402, adding more then one policy fails if you don't use python2.

vadmeste:
dict.iteritems() is not supported in python 3 but we can safely use
dict.items() which provides a _view_ of dict elements.

Fixes #585

### DIFF
--- a/minio/policy.py
+++ b/minio/policy.py
@@ -378,7 +378,7 @@ def _remove_statements(statements, policy, bucket_name, prefix=''):
 #
 def _merge_dict(d1, d2):
     out = collections.defaultdict(list)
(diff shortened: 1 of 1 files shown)


vadmeste:
> fold_case_dict.py also uses iteritems() method. That could be fixed as well.

iteritems() method in fold_case_dict doesn't refer to dict.iteritems() but it does refer to fold_case_dict.iteritems() which is an internal function.

poornas:
you are right @vadmeste.

Issue #200, opened 2015-07-10
Non-recursive list objects should return found common prefixes


fkautz:
Expected:
Return both objects and common prefixes

Actual:
Returns only objects

fkautz:
Implemented by #203

PR #587, opened 2017-10-26
Fixes gcs/azure bucket policy test issues

ebozduman:
#200 mint issue

### DIFF
--- a/minio/policy.py
+++ b/minio/policy.py
@@ -240,8 +240,8 @@ def _remove_bucket_actions(statement, policy, prefix=''):
 
     actions = _get_action(statement)
     cond_value = statement.get('Condition', {})
-    if (statement.get('Effect') == 'Allow' and
-            statement.get('Principal') == {'AWS': '*'}):
+    aws = statement.get('Principal', {}).get('AWS', [])
+    if (statement.get('Effect') == 'Allow' and (aws == '*' or '*' in aws) ):
         if policy == Policy.READ_ONLY:
             statement, actions, cond_value = _remove_read_only(statement,
                                                                actions,
@@ -265,9 +265,9 @@ def _remove_bucket_actions(statement, policy, prefix=''):
 # Removes object actions in given statement for given policy.
 def _remove_object_actions(statement, policy):
     actions = _get_action(statement)
-    if (statement.get('Effect') == 'Allow' and
-            statement.get('Principal') == {'AWS': '*'} and
-            not statement.get('Condition')):
+    aws = statement.get('Principal', {}).get('AWS', [])
+    if (statement.get('Effect') == 'Allow' and (aws == '*' or '*' in aws) and
+        not statement.get('Condition') ):
         if policy == Policy.READ_ONLY:
             actions = list(set(actions) - _READ_ONLY_OBJECT_ACTIONS)
         elif policy == Policy.WRITE_ONLY:
@@ -311,10 +311,9 @@ def _remove_statements(statements, policy, bucket_name, prefix=''):
             s = _remove_object_actions(s, policy)
 
         if s:
-            if (resource == bucket_resource and
-                    set(_get_action(s)) & _READ_ONLY_BUCKET_ACTIONS and
-                    s.get('Effect') == 'Allow' and
-                    s.get('Principal') == {'AWS': '*'}):
+            aws = s.get('Principal', {}).get('AWS', [])
+            if (resource == bucket_resource and set(_get_action(s)) & _READ_ONLY_BUCKET_ACTIONS and
+                s.get('Effect') == 'Allow' and (aws == '*' or '*' in aws) ):
                 cond_value = s.get('Condition', {})
                 string_equals_value = cond_value.get('StringEquals', {})
                 values = string_equals_value.get('s3:prefix', [])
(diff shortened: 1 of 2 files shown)


minio/policy.py
```diff
@@ -240,8 +240,10 @@ def _remove_write_only(actions, cond_value):
 
     actions = _get_action(statement)
     cond_value = statement.get('Condition', {})
+    principal = statement.get('Principal')
     if (statement.get('Effect') == 'Allow' and
-            statement.get('Principal') == {'AWS': '*'}):
+            (principal == {'AWS': '*'} or
+            '*' in principal['AWS'])):
```
balamurugana: This comparison is not very clean.  one check is dict match another one is string match.  We could avoid exact dict match.  You can do 
```py
aws = statement.get('Principal', {}).get("AWS", [])
```

then
```py
if (... and ('*' in aws or '*' in aws)):
```

tests/functional/tests.py
```diff
@@ -588,26 +588,72 @@ def test_get_bucket_policy(client, log_output):
...
         # Set read-only policy successfully.
         client.set_bucket_policy(bucket_name, prefix, Policy.READ_ONLY)
+        # Validate if the policy is set correctly
+        policy_name = client.get_bucket_policy(bucket_name)
+        if policy_name != Policy.READ_ONLY:
+            raise ValueError('Failed to set ReadOnly bucket policy: ' + policy_name)
+    except APINotImplemented:
+        print(log_output.json_report(alert='Not Implemented', status=LogOutput.NA))
+    except Exception as err:
+        raise Exception(err)
```
poornas: shoul;dn't this exception be captured in the log as well?

tests/functional/tests.py
```diff
@@ -588,26 +588,72 @@ def test_get_bucket_policy(client, log_output):
...
         # Set read-write policy successfully.
         client.set_bucket_policy(bucket_name, prefix, Policy.READ_WRITE)
+        # Validate if the policy is set correctly
+        policy_name = client.get_bucket_policy(bucket_name)
+        if policy_name != Policy.READ_WRITE:
+            raise ValueError('Failed to set ReadWrite bucket policy: ' + policy_name)
+    except APINotImplemented:
+        print(log_output.json_report(alert='Not Implemented', status=LogOutput.NA))
+    except Exception as err:
+        raise Exception(err)
```
poornas: exception needs to be captured to log..

tests/functional/tests.py
```diff
@@ -588,26 +588,72 @@ def test_get_bucket_policy(client, log_output):
         except Exception as err:
             raise Exception(err)
 
-def test_set_bucket_policy(client, log_output):
+def test_set_bucket_policy_readonly(client, log_output):
     # Get a unique bucket_name
     log_output.args['bucket_name'] = bucket_name = generate_bucket_name()
-    log_output.args['prefix'] = prefix = '1/'
+    log_output.args['prefix'] = prefix = ''
```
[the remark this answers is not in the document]
ebozduman: That is because gcs and azure do not support prefixes. I wanted to have the common ground for all gateway test cases.

tests/functional/tests.py
ebozduman: All of the generic exceptions like this one are handled and logged in the main() function.

tests/functional/tests.py
ebozduman: The same here.

minio/policy.py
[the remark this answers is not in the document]
ebozduman: Done.

ebozduman: Done.

minio/policy.py
```diff
@@ -240,8 +240,8 @@ def _remove_write_only(actions, cond_value):
 
     actions = _get_action(statement)
     cond_value = statement.get('Condition', {})
-    if (statement.get('Effect') == 'Allow' and
-            statement.get('Principal') == {'AWS': '*'}):
+    aws = statement.get('Principal', {}).get('AWS', [])
+    if (statement.get('Effect') == 'Allow' and (aws == '*' or '*' in aws) ):
```
poornas: ```(aws == '*' or '*' in aws) ```can be replaced by ```('*' in aws)``` everywhere this comparison is done.

minio/policy.py
```diff
@@ -311,10 +311,9 @@ def _remove_statements(statements, policy, bucket_name, prefix=''):
             s = _remove_object_actions(s, policy)
 
         if s:
-            if (resource == bucket_resource and
-                    set(_get_action(s)) & _READ_ONLY_BUCKET_ACTIONS and
-                    s.get('Effect') == 'Allow' and
-                    s.get('Principal') == {'AWS': '*'}):
+            aws = s.get('Principal', {}).get('AWS', [])
+            if (resource == bucket_resource and set(_get_action(s)) & _READ_ONLY_BUCKET_ACTIONS and
+                s.get('Effect') == 'Allow' and (aws == '*' or '*' in aws) ):
```
poornas: same comment as above

minio/policy.py
ebozduman: Our requirement is to get `True` for cases like `aws` is a string and strictly set to `'*'`, or `aws` is a list and at least one of its element is exactly`'*'`, like`['abc, ..., '*', ..., 'def']`. So, we need to reject or return `False` for string values like `'abc*'`.
If we drop `aws == '*'` part of the condition, then `('*' in aws)` will return `True` for cases like `'abc*'`, and we don't want that. That's why we have to have both of the conditions, `(aws == '*' or '*' in aws)`, in place.

poornas: ok, makes sense.

deekoder:
All changes requested are addressed. 2 approvals received. Had discussion with @ebozduman  and cleared up all questions. Merging hence.

PR #589, opened 2017-11-01
Allow users to use custom http client in the constructor

### LINKED ISSUE #588: Support for HTTP Proxies
While minio is usually deployed on-premise and the connection between client and server doesn't need to traverse HTTP Proxies, in large infrastructure setups or distributed ones between VPCs, the need for traversing proxies is real.

This could be achieved simply by using ProxyManager instead of PoolManager.

PR coming :)

kannappanr:
Minio constructor has been modified to take a custom http client
from the user.

This prevents the constructor from having too many options.

The next version has been bumped to 3.0.0, since this change
breaks backward compatibility.

Fixes #588

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -44,11 +44,11 @@ s3Client = Minio('s3.amazonaws.com',
 ## 1. Constructor
 
 <a name="Minio"></a>
-### Minio(endpoint, access_key=None, secret_key=None, secure=True)
+### Minio(endpoint, access_key=None, secret_key=None, secure=True, region=None, http_client=None)
 
 |   |
 |---|
-| `Minio(endpoint, access_key=None, secret_key=None, secure=True, region=None, timeout=None)`  |
+| `Minio(endpoint, access_key=None, secret_key=None, secure=True, region=None, http_client=None)`  |
 | Initializes a new client object.  |
 
 __Parameters__
@@ -61,7 +61,7 @@ __Parameters__
 | `secret_key` | _string_  |  Secret key for the object storage endpoint. (Optional if you need anonymous access). |
 | `secure`  |_bool_   | Set this value to `True` to enable secure (HTTPS) access. (Optional defaults to `True`).  |
 | `region`  |_string_ | Set this value to override automatic bucket location discovery. (Optional defaults to `None`). |
-| `timeout` |_float_ | Set this value to control how long requests are allowed to run before being aborted. (Optional defaults to `None`) |
+| `http_client` |_urllib3.poolmanager.PoolManager_ | Set this value to use custom http client instead of using default http client. (Optional defaults to `None`) |
 
 __Example__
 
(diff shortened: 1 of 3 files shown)


poornas:
@kannappanr, the minio docs needs to have documentation on the certificate_bundle and proxy_manager as optional args to the constructor.

kannappanr:
@poornas @balamurugana , re-uploaded the code and updated the document based on comments from @harshavardhana

minio/api.py
```diff
@@ -128,13 +128,15 @@ class Minio(object):
 
     def __init__(self, endpoint, access_key=None,
                  secret_key=None, secure=True,
-                 region=None,
-                 timeout=None,
-                 certificate_bundle=certifi.where()):
+                 region=None, http_client=None):
```
harshavardhana: Is it possible to do this instead? 

```
default_pool_manager = urllib3.poolmanager.PoolManager()

def __init__ (self, endpoint, access_key=None, 
                     secret_key=None, secure=True,
                     region=None, http_client=default_pool_manager):
```

This would avoid the `if` checks i think.

poornas:
build checks fail

kannappanr:
@poornas @balamurugana Build failures have been fixed. Ready for review.

docs/API.md
```diff
@@ -44,11 +44,11 @@ s3Client = Minio('s3.amazonaws.com',
 ## 1. Constructor
 
 <a name="Minio"></a>
-### Minio(endpoint, access_key=None, secret_key=None, secure=True)
+### Minio(endpoint, access_key=None, secret_key=None, secure=True, region=None, http_client=None)
 
 |   |
 |---|
-| `Minio(endpoint, access_key=None, secret_key=None, secure=True, region=None, timeout=None)`  |
+| `Minio(endpoint, access_key=None, secret_key=None, secure=True, region=None, http_client=None)`  |
```
balamurugana: We are breaking backward compatibility.  Is it expected?

minio/helpers.py
```diff
@@ -271,6 +274,17 @@ def get_target_url(endpoint_url, bucket_name=None, object_name=None,
...
+    """
+    Verify if endpoint is valid urllib3 http client.
+
+    :type client: urllib3.poolmanager.PoolManager
+    :return: True if the client is valid, Raise :exc:`InvalidArgumentError`
+       otherwise.
+    """
+    if isinstance(client, urllib3.poolmanager.PoolManager):
+        return True
+    raise InvalidArgumentError('HTTP client should be of instance `urllib3.poolmanager.PoolManager`')
```
balamurugana: This is not pythonic.  you throw an exception on `isinstance()` failure and return nothing.  (There is no nothing in python, you get `None`).  IMO this as separate function is not needed.

docs/API.md
harshavardhana: Yes @balamurugana - this is get python in line with minio-java

balamurugana: then the next version needs to be updated in this PR

minio/api.py
```diff
@@ -156,19 +159,19 @@ def __init__(self, endpoint, access_key=None,
...
-
-        self._http = urllib3.PoolManager(
-            timeout=self._conn_timeout,
-            cert_reqs='CERT_REQUIRED',
-            ca_certs=certificate_bundle,
-            retries=urllib3.Retry(
-                total=5,
-                backoff_factor=0.2,
-                status_forcelist=[500, 502, 503, 504]
+        if http_client is None:
```
balamurugana: you can do simply `if not http_client:`

minio/__init__.py
```diff
@@ -29,7 +29,7 @@
 
 __title__ = 'minio-py'
 __author__ = 'Minio, Inc.'
-__version__ = '2.2.7'
+__version__ = '3.0.0'
```
harshavardhana: :+1:

kannappanr:
@balamurugana and @poornas can you please review again?

PR #592, opened 2017-11-02
add parameter to skip SSL cert verification

feifangit:
when minio servers is set up with self-signed cert. we may set this new parameter `verify` to `False` (just like the `requests` library) to skip the verification. Otherwise, a `SSLError` would be thrown.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -130,7 +130,8 @@ class Minio(object):
                  secret_key=None, secure=True,
                  region=None,
                  timeout=None,
-                 certificate_bundle=certifi.where()):
+                 certificate_bundle=certifi.where(),
+                 verify=True):
 
         # Validate endpoint.
         is_valid_endpoint(endpoint)
@@ -161,7 +162,7 @@ class Minio(object):
 
         self._http = urllib3.PoolManager(
             timeout=self._conn_timeout,
-            cert_reqs='CERT_REQUIRED',
+            cert_reqs='CERT_REQUIRED' if verify else 'CERT_NONE',
             ca_certs=certificate_bundle,
             retries=urllib3.Retry(
                 total=5,
--- a/minio/api.py
+++ b/minio/api.py
@@ -123,6 +123,7 @@ class Minio(object):
          location discovery.
     :param timeout: Set this value to control how long requests
          are allowed to run before being aborted.
+    :param verify: Set this to `True` to skip SSL Cert Verification
     :return: :class:`Minio <Minio>` object
     """
 


harshavardhana:
We should avoid adding more fields to constructor, lets actually provide a function to add custom urllib3 instance. @feifangit

feifangit:
since minio py is using urllib3 underneath, you could provide a `**urllib_kwargs` in constructor  to `.update()` existing parameters. 

```python
        self._http = urllib3.PoolManager(
            timeout=self._conn_timeout,
            cert_reqs='CERT_REQUIRED',
            ca_certs=certificate_bundle,
            retries=urllib3.Retry(
                total=5,
                backoff_factor=0.2,
                status_forcelist=[500, 502, 503, 504]
            )
        )
```
to 
```
urllib3_kwargs = {'timeout': xxxx, 'cert_reqs': 'CERT_REQUIRED'....}
urllib3_kwargs.update(urllib_kwargs)
self._http = urllib3.PoolManager(urllib3_kwargs)
```

Ask users to pass a urllib3 instance is not a good idea, user needs to know more about `urllib3` even if they needs a simple connection. unless you abstract the entire transport layer, just like `celery` utilizes `kombu`.

harshavardhana:
> since minio py is using urllib3 underneath, you could provide a **urllib_kwargs in constructor to .update() existing parameters.
> 
>         self._http = urllib3.PoolManager(
>             timeout=self._conn_timeout,
>             cert_reqs='CERT_REQUIRED',
>             ca_certs=certificate_bundle,
>             retries=urllib3.Retry(
>                 total=5,
>                 backoff_factor=0.2,
>                 status_forcelist=[500, 502, 503, 504]
>             )
>         )
> to
> 
> urllib3_kwargs = {'timeout': xxxx, 'cert_reqs': 'CERT_REQUIRED'....}
> urllib3_kwargs.update(urllib_kwargs)
> self._http = urllib3.PoolManager(urllib3_kwargs)
> Ask users to pass a urllib3 instance is not a good idea, user needs to know more about urllib3 even if they needs a simple connection. unless you abstract the entire transport layer, just like celery utilizes kombu.

This is in line with what our other SDKs do, since you are interested in turning off cert verification you need to know what urllib3 provides, we are not doing anything really on our end.  

This way you can even provide your own instance such as proxy manager etc.

harshavardhana:
Also minio client constructor shouldn't look like a full blown constructor for a HTTP client it is only meaningful with limited context.  You should bring your own on http client if you wish to extend it.  So we will implement a method for you to pass in urllib3 instance instead.

nitisht:
@feifangit does #589 address your issue? if yes, feel free to close this PR

PR #599, opened 2017-12-07
Cleanup incomplete upload upon error

harshavardhana:
Since we do not have resuming support, we should
simply cleanup failed transfers.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -1527,7 +1527,15 @@ class Minio(object):
             parts_to_upload.append((bucket_name, object_name, upload_id, part_number, part_data))
 
         # Run parts upload in parallel
-        pool.parallel_run(self._upload_part_routine, parts_to_upload)
+        try:
+            pool.parallel_run(self._upload_part_routine, parts_to_upload)
+        except:
+            # Any exception that occurs sends an abort on the
+            # on-going multipart operation.
+            self._remove_incomplete_upload(bucket_name,
+                                           object_name,
+                                           upload_id)
+            raise
 
         # Update uploaded_parts with the part uploads result
         # and check total uploaded data.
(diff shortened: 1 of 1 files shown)


minio/api.py
```diff
@@ -1524,7 +1524,8 @@ def _stream_put_object(self, bucket_name, object_name,
 
             part_data = read_full(data, current_part_size)
             # Append current part information
-            parts_to_upload.append((bucket_name, object_name, upload_id, part_number, part_data))
+            parts_to_upload.append((bucket_name, object_name, upload_id,
```
kannappanr: Is this just a formatting change?

minio/api.py
```diff
@@ -1524,7 +1524,8 @@ def _stream_put_object(self, bucket_name, object_name,
...
+                                    part_number, part_data))
 
         # Run parts upload in parallel
         pool.parallel_run(self._upload_part_routine, parts_to_upload)
```
vadmeste: pool.parallel_run() can raise an exception, I think it would be good to catch it, remove incomplete upload and throw the exception again.

```
try:
     pool.parallel_run(self._upload_part_routine, parts_to_upload)
except:
     self._remove_incomplete_upload(bucket_name,
                                           object_name,
                                           upload_id)
     raise
```

harshavardhana: done

minio/api.py
harshavardhana: yep!

minio/api.py
```diff
@@ -1546,14 +1554,27 @@ def _stream_put_object(self, bucket_name, object_name,
...
+                                                           upload_id,
+                                                           uploaded_parts,
+                                                           metadata=metadata)
+        except:
+            # Any exception that occurs sends an abort on the
+            # on-going multipart operation.
+            self._remove_incomplete_upload(bucket_name,
+                                           object_name,
+                                           upload_id)
+
```
vadmeste: A missing raise here.

harshavardhana: done

minio/api.py
```diff
@@ -1546,14 +1554,28 @@ def _stream_put_object(self, bucket_name, object_name,
...
+                                           upload_id)
             raise InvalidSizeError(msg)
 
         # Complete all multipart transactions if possible.
-        mpart_result = self._complete_multipart_upload(bucket_name,
-                                                       object_name,
-                                                       upload_id,
-                                                       uploaded_parts,
-                                                       metadata=metadata)
+        try:
```
kannappanr: Looks like there is formatting issue here

harshavardhana: why do you think?

harshavardhana: i don't see any issue, if there is a formatting issue python won't really compile.

PR #609, opened 2018-01-10
Adds missing tests and fixes tests with problems

### LINKED ISSUE #562: Missing functional tests for Mint category
- [x] 1. Missing test for makebucket() with default region
- [x] 2. Missing test for makebucket() with region
- [x] 3. Missing test for makebucket() with bucket name with '.'
- [x] 4. Missing test for bucket_exists()
- [x] 5. Missing test for remove_bucket()
- [x] 6. Missing test for fput_object() with content type.
- [x] 7. Missing test for putobject() with content type.
- [x] 8. Missing test for putobject() multipart resume (may be missing feature)
- [x] 9. Missing test for putobject() with object name with path segment (may be not relevant)
- [x] 10. Missing test for putobject() for unknown sized stream for single put.
- [x] 11. Missing test for putobject() for unknown sized stream for multipart put.
- [x] 12. Missing test for putobject() for client side encryption.
- [x] 13. Missing test for stat_object()
- [x] 14. Missing test for getobject() with offset
- [x] 15. Missing test for getobject() with offset and length
- [x] 16. Missing test for getobject() object name with multiple path segment (may be not relevant)
- [x] 17. Missing test for getobject() with client side encryption.
- [x] 18. Missing test for listobject() with prefix and recursive
- [x] 19. Missing test for listobject() listing > 1000
- [x] 20. Missing test for listincompleteupload() with default/prefix and recursive (may be not relevant)
- [x] 21. Missing test for removeincompleteupload() (may be not relevant)
- [x] 22. Missing tests for presigned_getobject() with default timeout/specified timeout and response header.
- [x] 23. Missing tests for presigned_putobject() with default timeout and specified timeout.
- [x] 24. Missing test for thread safey of minio client object (may be not relevant)
- [x] 25. Missing test for copyobject() with conditions etag positive/negative cases and modified positive/negative cases.
- [x] 26. full tests and fast tests missing.

ebozduman:
Fixes #562

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -1208,7 +1208,6 @@ class Minio(object):
             if object_name == upload.object_name:
                 self._remove_incomplete_upload(bucket_name, object_name,
                                                upload.upload_id)
-                return
 
     def presigned_url(self, method,
                     bucket_name,
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -29,11 +29,14 @@ from datetime import datetime, timedelta
 
 import urllib3
 import certifi
+import hashlib
+from threading import Thread
 
 from minio import Minio, PostPolicy, CopyConditions
 from minio.policy import Policy
-from minio.error import (APINotImplemented, ResponseError, PreconditionFailed,
-                         BucketAlreadyOwnedByYou, BucketAlreadyExists)
+from minio.error import (APINotImplemented, ResponseError,
+                         PreconditionFailed, BucketAlreadyOwnedByYou,
+                         BucketAlreadyExists, InvalidBucketError)
 
 class LimitedRandomReader(object):
     """
@@ -52,11 +55,9 @@ class LimitedRandomReader(object):
     def __init__(self, limit):
         self._limit = limit
         self._offset_location = 0
-
     def read(self, amt=64*1024):
         """
         Similar to :meth:`io.read`, with amt option.
-
         :param amt:
             How much of the content to read.
         """
@@ -64,7 +65,6 @@ class LimitedRandomReader(object):
         if self._offset_location == self._limit:
             # return empty bytes to indicate EOF.
             return b''
-
         # make translation table from 0..255 to 97..122
         bal = [c.encode('ascii') for c in ascii_lowercase]
         amt = min(amt, self._limit - self._offset_location)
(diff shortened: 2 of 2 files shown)


minio/api.py
```diff
@@ -1208,7 +1208,6 @@ def remove_incomplete_upload(self, bucket_name, object_name):
             if object_name == upload.object_name:
                 self._remove_incomplete_upload(bucket_name, object_name,
                                                upload.upload_id)
-                return
```
vadmeste: Good catch!

tests/functional/tests.py
```diff
@@ -145,40 +145,103 @@ def json_report(self, err_msg='', alert='', status=''):
...
-
-    is_s3 = client._endpoint_url.startswith("s3.amazonaws")
+    # Default location
+    log_output.args['location'] = "default value ('us-east-1')"
     try:
-        # Create a bucket
+        # Create a bucket with default bucket location
         client.make_bucket(bucket_name)
         # Check if bucket was created properly
+        log_output.function = 'bucket_exists(bucket_name)'
```
vadmeste: Isn't it little confusing to change function name multiple times in the same test function ? I thought log_output.function should represent the name of the function.

kannappanr: It is the name of the API being tested

ebozduman: If you check the end of `tests.py` file (starting with line#1555), you notice, we first create an instance of `LogOutput` class, `log_output`, and we initialize the `function` attribute with the name of the api command/method that will go under test.

For example, in the above test case `test_make_bucket_default_region`, `log_output.function` is set to `make_bucket(bucket_name, location)` by default.

Because some tests, like `test_make_bucket_default_region`, depend on some other `helper` api commands, it would be misleading to report a helper command ,`bucket_exists`, failure with the name of the api command, `make_bucket`, under test. That's why, I  set `log_output.function` to 'bucket_exists(bucket_name)'` before calling `bucket_exists` api/helper command. The same with `remove_bucket` helper api command. This way, the name of the failed api command  will be reflected in the generated Json format output in the `function` field.

Since the code is not clear enough, (@vadmeste and I exchanged ideas and we agreed on) it would be helpful to add comments that explain the default value of `log_output.function` in each test definition.

ebozduman: Done.

vadmeste: we can simplify the code in another PR

PR #621, opened 2018-01-29
any specific url expiry positive time, default time is 7 days

alienr:
removing fixed url expiry time, making it any positive time.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -1240,7 +1240,7 @@ class Minio(object):
         is_non_empty_string(object_name)
 
         if expires.total_seconds() <= 0:
-            raise InvalidArgumentError('Expires param must be positive value)
+            raise InvalidArgumentError('Expires param must be positive value')
 
         region = self._get_bucket_region(bucket_name)
         url = get_target_url(self._endpoint_url,
--- a/minio/api.py
+++ b/minio/api.py
@@ -96,9 +96,6 @@ _DEFAULT_USER_AGENT = 'Minio {0} {1}'.format(
                      __version__))
 
 
-# Duration of 7 days in seconds
-_MAX_EXPIRY_TIME = 604800 # 7 days in seconds
-
 # Number of parallel workers which upload parts
 _PARALLEL_UPLOADERS = 3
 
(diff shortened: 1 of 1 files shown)


vadmeste:
Thanks for this PR @alienr,

Can you elaborate why this patch is needed ?

alienr:
thanks for your reply @vadmeste ,

I'm storing presigned urls into database, and If after 7 days it gets expired then I have to create the same again. so I was thinking about a positive expiry time only. there should be expiry of url as feature but it should not be a restriction.

vadmeste:
@alienr do you mean that you are using an S3 server that is not respecting 7 days maximum expiry of presigned urls ? I am not aware of any.

alienr:
@vadmeste Okay If I want to set expiry time 15 days, the current minio-py will not allow me.

alienr:
So I want flexible expiry time, not only 7 days.

harshavardhana:
S3 API mandates 7 days expiry. Minio server doesn't support 15 days URL expiries.  The current code is in accordance with S3 API requirements we don't need this PR.

alienr:
Hi, @harshavardhana 
the problem I'm facing is when I'm passing expiry time more than 7 days, minio-py throwing error (Expires param valid values are between 1 sec to 604800). so I'm not able to set greater expiry time than 7 days. AWS S3 allows any positive value as expiry time. 
Please explain why expiry time is restricted to 7 days only.

harshavardhana:
> the problem I'm facing is when I'm passing expiry time more than 7 days, minio-py throwing error (Expires param valid values are between 1 sec to 604800). so I'm not able to set greater expiry time than 7 days. AWS S3 allows any positive value as expiry time.
> Please explain why expiry time is restricted to 7 days only.

This is only true for AWS Signature V2, minio-py only implements Signature V4. https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html

`X-Amz-Expires` - Provides the time period, in seconds, for which the generated presigned URL is valid. For example, 86400 (24 hours). This value is an integer. The minimum value you can set is 1, and the maximum is 604800 (seven days). A presigned URL can be valid for a maximum of seven days because the signing key you use in signature calculation is valid for up to seven days.

kannappanr:
As explained by @harshavardhana , S3 signature V4 supports a maximum of 7 days for `X-Amz-Expires`. So, closing this PR as it violates S3 compatibility.

alienr:
thanks

PR #624, opened 2018-01-31
Override all other regions if region is set

### LINKED ISSUE #623: Hardcode region to 'us-east-1' for making bucket api
I am not sure that is a bug or intentional by hardcoding region to 'us-east-1' in making bucket api.
https://github.com/minio/minio-py/blob/188dba3dff087ef0d13e4d6336dcfcfc7c914f49/minio/api.py#L267

harshavardhana:
Fixes #623

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -264,7 +264,7 @@ class Minio(object):
         url = self._endpoint_url + '/' + bucket_name + '/'
 
         # Get signature headers if any.
-        headers = sign_v4(method, url, 'us-east-1',
+        headers = sign_v4(method, url, location,
                           headers, self._access_key,
                           self._secret_key, content_sha256_hex)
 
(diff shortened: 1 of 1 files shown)


harshavardhana:
> LGTM. Would be good candidate for a Mint test.

We cannot add a mint test because Minio server configuration or the choice of remote endpoint is not governed by mint.

minio/api.py
```diff
@@ -296,6 +296,9 @@ def list_buckets(self):
 
         # default for all requests.
         region = 'us-east-1'
+        # region is set then use the region.
```
ebozduman: Since we call the same code block twice, we can make it a simple method.

harshavardhana: There is no reason to make this a method @ebozduman - it will confuse with _get_bucket_location() - i will prefer to keep it as is.

PR #628, opened 2018-02-19
Change ThreadPool and Worker class in thread_pool.py

### LINKED ISSUE #595: python minio client may cause thread exhaust
- python: 3.5
- minio-client: 2.2.6
- minio-server info: 
VERSION 2017-10-27T18:59:02Z
MEMORY
Used: 7.5 MB | Allocated: 50 GB | Used-Heap: 7.5 MB | Allocated-Heap: 29 MB
PLATFORM
Host: 6cbd0abfd258 | OS: linux | Arch: amd64
RUNTIME
Version: go1.9.2 | CPUs: 12

I use python script to create img and save record into db, and use minio python client upload to minio server.
```python
minio_client = Minio('192.168.1.144:9000',
                     access_key='',
                     secret_key='',
                     secure=False)


def upload(file_path, key_name):
  try:
    minio_client.fput_object('intellid',
                             key_name,
                             file_path,
                             content_type='image/png',
                             metadata={'Content-Disposition': 'attachment'})
  except ResponseError as err:
    logging.exception(file_path)
    logging.exception('upload failed!')


def main():
  main_id = threading.get_ident()
  db = default_session()
  gens = db.query(Generation).filter(Generation.status == None).limit(5).all()

  processed = 0
  while gens and len(gens) != 0:
    for gen in gens:
      compositions = copy.deepcopy(gen.composition['compositions'])
      for composition in compositions:
        composition['img_path'] = path.join(data_home, path.basename(
          db.query(Img).filter(Img.guid == composition[
            'guid']).first().oss))

      out_img = compose(gen.composition['width'], gen.composition['height'],
                        gen.composition['linkType'], compositions)

      key_name = 'ori/' + str(uuid.uuid4()) + '.png'
      key_name_thumb = 'thumb/' + str(uuid.uuid4()) + '.png'

      tmp = path.join(prod_out, key_name)
      thumb_tmp = path.join(prod_out, key_name_thumb)

      out_img.save(tmp)
      res = post_proc(tmp, thumb_tmp, 300)

      guid = uuid.uuid1().int >> 65
      while True:
        img = db.query(Img).filter(Img.guid == guid).first()
        if not img:
          break
        guid = uuid.uuid1().int >> 65
      minio_oss = minio_oss_prefix + key_name
      minio_oss_thumb = minio_oss_prefix + key_name_thumb

      new_img = Img(guid=guid, oss=minio_oss, thumb_oss=minio_oss_thumb,
                    file_type='PNG', img_type='AI', attr=res)
      print("before upload acitve_thread {}.".format(threading.active_count()))
      upload(tmp, key_name)
      upload(thumb_tmp, key_name_thumb)
      print("after upload acitve_thread {}.".format(threading.active_count()))
      db.add(new_img)
      db.commit()
      gen.guid = guid
      gen.status = 0
      processed += 1
      db.commit()
      print("processed {}.".format(processed))
    # end for
    gens = db.query(Generation).filter(Generation.status == None).limit(
      5).all()
  # end while


if __name__ == '__main__':
  main()

```
> partial log print:
before upload acitve_thread 319.
after upload acitve_thread 322.
processed 107.
before upload acitve_thread 322.
after upload acitve_thread 325.
processed 108.
before upload acitve_thread 325.
after upload acitve_thread 328.
processed 109.

**problem:** the minio client will create `ThreadPool` with 3 thread (code location: minio/api.py line 1511). And these thread (minio.thread_pool.Worker) will nerver exit. After every iteration in loop, the active thread number will increase 3. Finally the total threads in current process will exceed the linux limit(2048 in `ulimit -a`).

I'm a novice in minio. appreciation for any help. thanks.

nnehdi:
Fixes #595

### DIFF
--- a/minio/thread_pool.py
+++ b/minio/thread_pool.py
@@ -83,7 +83,6 @@ class ThreadPool:
         if not self.exceptions_queue.empty():
             raise self.exceptions_queue.get()
 
-
     def result(self):
         """ Return the result of all called tasks """
         return self.results_queue
--- a/minio/thread_pool.py
+++ b/minio/thread_pool.py
@@ -37,12 +37,13 @@ class Worker(Thread):
         self.tasks_queue = tasks_queue
         self.results_queue = results_queue
         self.exceptions_queue = exceptions_queue
-        self.daemon = True
+        self.daemon = False
         self.start()
 
     def run(self):
         fast_quit = False
-        while True:
+
+        while not self.tasks_queue.empty():
             func, args, kargs = self.tasks_queue.get()
             if not fast_quit:
                 try:
@@ -61,8 +62,7 @@ class ThreadPool:
         self.results_queue = queue()
         self.exceptions_queue = queue()
         self.tasks_queue = queue(num_threads)
-        for _ in range(num_threads):
-            Worker(self.tasks_queue, self.results_queue, self.exceptions_queue)
+        self.num_threads = num_threads
 
     def add_task(self, func, *args, **kargs):
         """ Add a task to the queue """
@@ -72,6 +72,10 @@ class ThreadPool:
         """ Add a list of tasks to the queue """
         for args in args_list:
             self.add_task(func, args)
+            
+        for _ in range(self.num_threads):
+            Worker(self.tasks_queue, self.results_queue, self.exceptions_queue)
+
         # Wait for completion of all the tasks in the queue
         self.tasks_queue.join()
         # Check if one of the thread raised an exception, if yes
@@ -79,7 +83,9 @@ class ThreadPool:
         if not self.exceptions_queue.empty():
             raise self.exceptions_queue.get()
 
+
     def result(self):
         """ Return the result of all called tasks """
         return self.results_queue
+
 
--- a/minio/thread_pool.py
+++ b/minio/thread_pool.py
@@ -37,7 +37,7 @@ class Worker(Thread):
         self.tasks_queue = tasks_queue
         self.results_queue = results_queue
         self.exceptions_queue = exceptions_queue
-        self.daemon = False
+        self.daemon = True
         self.start()
 
     def run(self):
@@ -61,7 +61,7 @@ class ThreadPool:
     def __init__(self, num_threads):
         self.results_queue = queue()
         self.exceptions_queue = queue()
-        self.tasks_queue = queue(num_threads)
+        self.tasks_queue = queue()
         self.num_threads = num_threads
 
     def add_task(self, func, *args, **kargs):
--- a/minio/thread_pool.py
+++ b/minio/thread_pool.py
@@ -26,12 +26,13 @@ in parallel. Tasks results can also be retrieved.
 
 """
 
-import sys
 from threading import Thread
 from .compat import queue
 
+
 class Worker(Thread):
     """ Thread executing tasks from a given tasks queue """
+
     def __init__(self, tasks_queue, results_queue, exceptions_queue):
         Thread.__init__(self)
         self.tasks_queue = tasks_queue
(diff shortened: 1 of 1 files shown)


minio/thread_pool.py
```diff
@@ -61,8 +62,7 @@ def __init__(self, num_threads):
         self.results_queue = queue()
         self.exceptions_queue = queue()
         self.tasks_queue = queue(num_threads)
```
vadmeste: @nnehdi you need to remove queue size restriction here.. otherwise multipart PUT won't work. The reason is that the  code in this PR will fill tasks_queue with all tasks before starting to execute them.

minio/thread_pool.py
```diff
@@ -37,12 +37,13 @@ def __init__(self, tasks_queue, results_queue, exceptions_queue):
         self.tasks_queue = tasks_queue
         self.results_queue = results_queue
         self.exceptions_queue = exceptions_queue
-        self.daemon = True
+        self.daemon = False
```
ebozduman: Why do we turn off daemon mode, `self.daemon = False` ?

minio/thread_pool.py
ebozduman: I've tested the existing code (without @vadmeste's suggested correction above) and verified that multi-part uploads fails when more than `num_threads` parts are involved in the multi-part upload. That is, when the `tasks-queue` size is limited with `num_threads`, 3 in our case, the multi-part upload hangs with files bigger than 15MB (3 x 5MB - multi-part size limit) since `tasks_queue` gets full after 3rd part is `tasks_queue.put` in and the `tasks_queue.put` action for the 4th part blocks the operation. 
So, we should not limit the size of the `tasks-queue` :
`self.tasks_queue = queue(num_threads)` => `self.tasks_queue = queue()`

nnehdi: @ebozduman @vadmeste I agree with you, I did not consider that. As @ebozduman said we should remove the limit of queue size because we put all chunks of a file before we start the Worker threads.
thank you, guys.

minio/thread_pool.py
nnehdi: It depends on your situation. In our use case, we want to close the thread immediately after the program stopped and it's not important for us to miss an uploading file. I change this flag to its previous value (True) but It's better to have a config that user can change this flag depends on their situation when using py-mino.
tnx

ebozduman: @nnehdi,
Thank you very much for working on this fix.
Are you going to push these review comment changes, so that we can merge and close this PR?

minio/thread_pool.py
```diff
@@ -72,14 +72,20 @@ def parallel_run(self, func, args_list):
...
+            Worker(self.tasks_queue, self.results_queue, self.exceptions_queue)
+
         # Wait for completion of all the tasks in the queue
         self.tasks_queue.join()
         # Check if one of the thread raised an exception, if yes
         # raise it here in the function
         if not self.exceptions_queue.empty():
             raise self.exceptions_queue.get()
 
+
```
ebozduman: Could you also clean up these last 2 extra lines introduced?

ebozduman: Could you also clean up these last 2 extra space lines introduced?

minio/thread_pool.py
```diff
@@ -42,7 +42,8 @@ def __init__(self, tasks_queue, results_queue, exceptions_queue):
 
     def run(self):
         fast_quit = False
-        while True:
+
```
ebozduman: We don't need this extra line.

minio/thread_pool.py
```diff
@@ -83,3 +87,4 @@ def result(self):
         """ Return the result of all called tasks """
         return self.results_queue
 
+
```
ebozduman: We don't need this extra line.
Could you also squash when you are done.

PR #634, opened 2018-02-27
[WIP] Add support for progress in put functions

### LINKED ISSUE #601: I want see a progress bar
when i put some file using minio-py 
```sh
fput_object()
```
it is good for some small file but large file, it is take a long time while uploading.

so user's doesn't know it is in progress or not.

so could you add option like ```progress_use``` ?

thanks.

kannappanr:
Add support for progress notifier in both fput_object and
put_object

Fixes #601

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -748,6 +748,7 @@ __Parameters__
 |``length``   |_int_   |Total length of object.   |
 |``content_type``   |_string_ | Content type of the object. (optional, defaults to 'application/octet-stream').   |
 |``metadata``   |_dict_ | Any additional metadata. (optional, defaults to None). |
+|``progress``   |_io.BufferedRandom_ | Dummy reader to track the progress of the operation. (optional, defaults to None). |
 
 __Return Value__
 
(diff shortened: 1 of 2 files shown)


vadmeste:
@kannappanr the problem with the current style is that there is no "realtime" update of the progress (only b 5Mb). I don't think users will like it either.

What about wrapping data reader with new Write() to update progress for each Write() call.

I am not sure if it works, take a look on this https://stackoverflow.com/questions/4013843/how-to-wrap-file-object-read-and-write-operation-which-are-readonly

harshavardhana:
> @kannappanr the problem with the current style is that there is no "realtime" update of the progress (only b 5Mb). I don't think users will like it either.
> 
> What about wrapping data reader with new Write() to update progress for each Write() call.
> 
> I am not sure if it works, take a look on this https://stackoverflow.com/questions/4013843/how-to-wrap-file-object-read-and-write-operation-which-are-readonly

Yes we need to ask for a progress if provided should atleast have Read()/Write() implemented. Such that each read or write we will progress reader as well, look at the hook-reader implementation in minio-go which will give ideas on how to do it in streaming fashion.

A similar example we have in our functional tests which implements RandomReader with Limits.

harshavardhana:
@kannappanr are you still working on this?

nitisht:
@kannappanr are you planning to work on this

kannappanr:
@nitisht I will not be working on this soon. So, will close this for now.

PR #652, opened 2018-04-17
fix(api): json error handling

fdhadzh:


minio/api.py
```diff
@@ -552,7 +552,7 @@ def listen_bucket_notification(self, bucket_name, prefix='', suffix='',
                     event = json.loads(line)
                     if event['Records'] is not None:
                         yield event
-            except SyntaxError:
+            except json.decoder.JSONDecodeError:
```
ebozduman: Related comment above at line 524 also needs to be modified to reflect the change.
Maybe like:
`NOTE: Notification is retried in case of 'SyntaxError' otherwise`
 =>
`NOTE: Notification is retried in case of json syntax error, 'json.decoder.JSONDecodeError', otherwise`

nitisht:
@fdhadzh do you mind closing this as problem is now fixed by https://github.com/minio/minio-py/pull/657

PR #657, opened 2018-05-30
fix listen_bucket_notification

WaYdotNET:
use json.decoder.JSONDecodeError instead of SyntaxError

ref #652

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -39,6 +39,11 @@ import os
 import itertools
 import codecs
 
+try:
+    from json.decoder import JSONDecodeError
+except ImportError:
+    JSONDecodeError = ValueError
+
 # Dependencies
 import urllib3
 import certifi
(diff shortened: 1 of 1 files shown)


minio/api.py
```diff
@@ -505,7 +505,7 @@ def listen_bucket_notification(self, bucket_name, prefix='', suffix='',
                     event = json.loads(line)
                     if event['Records'] is not None:
                         yield event
-            except SyntaxError:
+            except json.decoder.JSONDecodeError:
```
vadmeste: @WaYdotNET thanks for the PR.

We actually want to support both python 2 & 3. Python 2 doesn't know json.decoder.JSONDecodeError and parsing wrong json raises ValueError exception. Now, since json.decoder.JSONDecodeError is a subclass of ValueError, catching ValueError instead of json.decoder.JSONDecodeError will work for both python 2 & 3.

What do you think ?

WaYdotNET: I prefered check if JSONDecodeError exist 
:
```python
try:
    from json.decoder import JSONDecodeError
except ImportError:
    JSONDecodeError = ValueError
```

```python
    except JSONDecodeError:
```
What do you think ?

vadmeste: @WaYdotNET I am good with your idea.

PR #658, opened 2018-06-01
tests: Negative tests should ignore exceptions properly

harshavardhana:


tests/functional/tests.py
```diff
@@ -628,12 +628,23 @@ def test_negative_put_object_with_path_segment(client, log_output):
                           object_name,
                           io.BytesIO(b''), 0)
     except ResponseError as err:
-        pass
+        if err.code != 'XMinioInvalidObjectName':
+            raise Exception(err)
```
balamurugana: It looks ugly to wrap an exception with `Exception` which is unnecessary.  I see this is used all over the code.

harshavardhana: Yes I kept it as it is.. will need to change comprehensively later.

tests/functional/tests.py
```diff
@@ -628,12 +628,23 @@ def test_negative_put_object_with_path_segment(client, log_output):
...
+        else:
+            pass
     except Exception as err:
         raise Exception(err)
     finally:
         try:
-            client.remove_object(bucket_name, object_name)
+            try:
```
ebozduman: There are 2 `try` blocks back to back.
First `try` block is unnecessary.

harshavardhana: you need it... otherwise, we will crash. Look at the original issue why this fix is necessary. #5954

tests/functional/tests.py
```diff
@@ -628,12 +628,23 @@ def test_negative_put_object_with_path_segment(client, log_output):
...
+                client.remove_object(bucket_name, object_name)
+            except ResponseError as err:
+                if err.code != 'XMinioInvalidObjectName':
+                    raise Exception(err)
+                else:
+                    pass
+            except Exception as err:
```
poornas: lines 646,647 can be removed -since they are just re-raising the exception that would have been raised even without these statements.

tests/functional/tests.py
```diff
@@ -628,12 +628,23 @@ def test_negative_put_object_with_path_segment(client, log_output):
                           object_name,
                           io.BytesIO(b''), 0)
     except ResponseError as err:
-        pass
+        if err.code != 'XMinioInvalidObjectName':
+            raise Exception(err)
+        else:
+            pass
     except Exception as err:
         raise Exception(err)
```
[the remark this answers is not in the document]
harshavardhana: I am not able to understand... how would that happen?

tests/functional/tests.py
harshavardhana: I am not able to understand... how would that happen?

PR #659, opened 2018-06-07
Avoiding continuous polling in listen_bucket_notification 

Praveenrajmani:
The server streams an empty string (" ") until an action hits the bucket, 
In this case `JSONEncodeError` exception is thrown for every `json.loads(" ")` , So the loop continues here and polls the request for every catch .

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -507,9 +507,10 @@ class Minio(object):
                                       query=query, preload_content=False)
             try:
                 for line in response.stream():
-                    event = json.loads(line)
-                    if event['Records'] is not None:
-                        yield event
+                    if line.strip():
+                        event = json.loads(line)
+                        if event['Records'] is not None:
+                            yield event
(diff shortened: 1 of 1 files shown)


minio/api.py
```diff
@@ -507,9 +507,10 @@ def listen_bucket_notification(self, bucket_name, prefix='', suffix='',
                                       query=query, preload_content=False)
             try:
                 for line in response.stream():
-                    event = json.loads(line)
-                    if event['Records'] is not None:
-                        yield event
+                    if len(line.strip()) > 0:
```
ebozduman: It is just a preference, but IMHO, `if not line.strip():` is better than checking the length of a string.

Praveenrajmani: yes @ebozduman this is much simpler. Changing it

Praveenrajmani: @ebozduman  Done

PR #670, opened 2018-06-20
Do not urldecode Key from list_objects response

### LINKED ISSUE #669: urldecode of key in list_objects
We have some filenames in our Bucket that contain a percent (%) sign. When using `mc`, they are display normally and can be accessed.
Using the python library, the file name (key) gets URL decoded (falsely) and the resulting key name cannot be accessed any more.

I tried to get Minio to actually send URL-encoded key names in a list_objects call (using spaces, umlauts or chinese characters in key names), but they are always sent UTF8-encoded inside the XML. 

So why is the key name urldecoded at all? Can this be removed?

DeKe42:
Fixes #669

### DIFF
--- a/minio/parsers.py
+++ b/minio/parsers.py
@@ -197,7 +197,7 @@ def _parse_objects_from_xml_elts(bucket_name, contents, common_prefixes):
     """
     objects = [
         Object(bucket_name,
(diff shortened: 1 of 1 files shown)


DeKe42:
Sure, I did not notice that. It's included now.

PR #678, opened 2018-07-12
Fix parse_new_multipart_upload

messense:
I believe it should be `InitiateMultipartUploadResult`

https://github.com/minio/minio/blob/3dc13323e51dc7038232f5f02f55b37b388c59c2/cmd/api-response.go#L228-L234


messense:
Ping.

PR #697, opened 2018-09-13
Handle empty XML response in get_bucket_notification

### LINKED ISSUE #696: get_bucket_notification example throws minio.error.InvalidXMLError
using the play.minio.io server and the examples from the Python Client API Reference

```
from minio import Minio

minioClient = Minio('play.minio.io:9000',
                  access_key='Q3AM3UQ867SPQQA43P2F',
                  secret_key='zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG',
                  secure=True)

buckets = minioClient.list_buckets()
for bucket in buckets:
    print(bucket.name, bucket.creation_date)
    notification = minioClient.get_bucket_notification(bucket.name)
    break
```

throws the following error:
```
  File "<stdin>", line 3, in <module>
  File "/python3.6/site-packages/minio/api.py", line 424, in get_bucket_notification
    return parse_get_bucket_notification(data)
  File "/python3.6/site-packages/minio/parsers.py", line 365, in parse_get_bucket_notification
    root = S3Element.fromstring('GetBucketNotificationResult', data)
  File "/python3.6/site-packages/minio/parsers.py", line 76, in fromstring
    root_name, error.message
minio.error.InvalidXMLError: InvalidXMLError: message: "GetBucketNotificationResult" XML is not parsable. Message: no element found: line 1, column 0
```

krisis:
- Fixes #696 
- Added functional test

### DIFF
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -1484,6 +1484,32 @@ def policy_validated(client, bucket_name, policy):
         return False
     return True
 
+def test_get_bucket_notification(client, log_output):
+    # Get a unique bucket_name
+    log_output.args['bucket_name'] = bucket_name = generate_bucket_name()
+    try:
+        # make bucket as a preparation for empty bucket notification
+        client.make_bucket(bucket_name)
+        notification = client.get_bucket_notification(bucket_name)
+        if notification != {}:
+            raise ValueError("Failed to receive an empty bucket notification")
+
+    except APINotImplemented:
+            print(log_output.json_report(alert='Not Implemented',
+                                         status=LogOutput.NA))
+    except Exception as err:
+            print("exception", err)
+            raise Exception(err)
+    else:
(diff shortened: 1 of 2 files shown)


tests/functional/tests.py
```diff
@@ -1910,6 +1936,10 @@ def main():
             log_output =  LogOutput(client.set_bucket_policy, 'test_set_bucket_policy_readonly')
             test_set_bucket_policy_readonly(client, log_output)
 
+            log_output = LogOutput(client.get_bucket_notification, 'test_get_bucket_notification')
```
ebozduman: It looks like the test has been added only into `QuickMode`.
We also  need to add the same test into `FullMode`.

krisis: @ebozduman I have added the test to full mode too.

PR #698, opened 2018-09-17
minio-py should honour `is_dir` in listobjects correctly

### LINKED ISSUE #695: list_objects never populating is_dir=True?
I'm trying to understand the `is_dir` key of the objects returned by `client.list_objects()` and `client.list_objects_v2()`.

I am only ever seeing the value set to `False`

```
<Object: bucket_name: testingbucket object_name: b'thefolder/' last_modified: 2018-09-05 00:46:46+00:00 etag: d41d8cd98f00b204e9800998ecf8427e size: 0 content_type: None, is_dir: False, metadata: None>
```

I was expecting to see this folder reporting with `is_dir: True`.

I am testing with AWS S3 proper.  Does that matter?

Am I holding it wrong?

Praveenrajmani:
is_dir for listobjects was always set to `False` by default
So, it should be handled properly by decisions on the object_name , 
ie, to consider it as a dir if the name has "/" suffixed.
 
Fixes #695

### DIFF
--- a/minio/parsers.py
+++ b/minio/parsers.py
@@ -143,6 +143,14 @@ class S3Element(object):
         """
         return self.element.text
 
+    def is_dir(self):
+        """Returns True if the object is a dir
+        ie, if an object name has `/` suffixed.
+
+        """
+        text = self.get_child_text('Key')
+        return text.endswith("/")
+
 
 def parse_multipart_upload_result(data):
     """
@@ -200,7 +208,8 @@ def _parse_objects_from_xml_elts(bucket_name, contents, common_prefixes):
                content.get_child_text('Key'),
                content.get_localized_time_elem('LastModified'),
                content.get_etag_elem(strict=False),
-               content.get_int_elem('Size'))
+               content.get_int_elem('Size'),
+               is_dir=content.is_dir())
         for content in contents
     ]
 


minio/parsers.py
```diff
@@ -143,6 +143,14 @@ def text(self):
...
         return self.element.text
 
+    def is_dir(self):
+        """Returns True if the object is a dir
+        ie, if an object name has `/` suffixed.
+
+        """
+        text = self.get_child_text('Key')
+        return text.endswith("/")
+
```
nitisht: can you add a functional test for this as well

Praveenrajmani: yes .. ok

Praveenrajmani: Seems minio-py does not allow directory/empty puts , ie, we cannot do `mb play/test/testb` or `put obj/metaobj/` as per minio-py standards. 
AFAIK such puts can be done `mc` (`mc mb test/b/metab`) or `awscli`/`boto3`. So this test needs to be in mc/awscli. What say @nitisht ?

Praveenrajmani: Or adding unit test is fine ?

Praveenrajmani:
Since minio does not support object names with trailing `/`, we cannot add tests. So leaving it as is.

PR #700, opened 2018-09-21
Set Encryption headers only for SSE-C in multipart uploads

kannappanr:
Encryption headers are not required during the PUT object part API call when creating an object using SSE-S3 and SSE-KMS encryption options. AWS S3 fails if these headers are set in this scenario.

Fix is same as minio/minio-go#1026

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -1459,15 +1459,18 @@ class Minio(object):
         if metadata:
             headers.update(metadata)
 
-        if sse:
-            headers.update(sse.marshal())
-
         query = {}
         if part_number > 0 and upload_id:
             query = {
                 'uploadId': upload_id,
                 'partNumber': str(part_number),
             }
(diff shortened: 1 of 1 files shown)


minio/api.py
```diff
@@ -1459,15 +1459,18 @@ def _do_put_object(self, bucket_name, object_name, part_data,
...
-
         query = {}
         if part_number > 0 and upload_id:
             query = {
                 'uploadId': upload_id,
                 'partNumber': str(part_number),
             }
+            # Encryption headers for multipart uploads should
+            # be set only in the case of SSE-C.
+            if sse and type(sse).__name__ == "SSE-C":
```
poornas: @kannappanr,  instead of ```type(sse).__name__```, you can check ```sse.type()```

poornas:
travis failure needs to be addressed

PR #703, opened 2018-09-29
multipart: Avoid reading all parts at once before upload

### LINKED ISSUE #626: Stream multipart PUT operation
when i put huge file using minio-py
```
_stream_put_object
```

it is good for some small file but not so good large file, since it is read the entire data-stream as partition park into memory before the parallel upload, and could be an issue for running the application in small instance.
screen shot 2018-02-13 at 2 24 22 pm

Shall it only read the data stream partition (part-by-part) while uploading? it will reduce unnecessary  memory piling-up in application that queuing to be sent.

So, what i did is following:
1. Override _stream_put_object  
2. De-couple (remove) the parts reading/generating
3. Make it as a generator that return the part_data in batch, the parallel_run will need to call it every-time to return the next value

Read data in batch
```
# -- Batch Data Read --
def data_batch(bucket_name, object_name, upload_id,
               data, total_parts_count, part_size, last_part_size):
    """
    Read Raw Data by Batch
    :param bucket_name: Bucket to retrieve object from
    :param object_name: Name of object to retrieve
    :param upload_id: Unique Id for upload
    :param data: Raw data request stream
    :param total_parts_count: Number of part
    :param part_size: Size of each part
    :param last_part_size:  Size of last part
    :return: part number, part_data
    """
    for part_number in range(1, total_parts_count + 1):
        current_part_size = (part_size if part_number < total_parts_count
                             else last_part_size)
        part_data = read_full(data, current_part_size)
        yield bucket_name, object_name, upload_id, part_number, part_data
```

On the original code just call the existing parallel_run and pass it the data batch function
```
        # Stream parts upload in parallel <= current_part_size until
        # part_number reaches total_parts_count calculated for the
        # given size. Additionally part_manager() also provides
        # md5digest and sha256digest for the partitioned data.
        try:
            pool.parallel_run(self._upload_part_routine,
                              data_batch(bucket_name, object_name, upload_id,
                                         data, total_parts_count,
                                         part_size, last_part_size))
        except:
            # Any exception that occurs sends an abort on the
            # on-going multipart operation.
            self._remove_incomplete_upload(bucket_name,
                                           object_name,
                                           upload_id)
            raise
```

Does it make sense? it also could useful to others.  Your input and feedback would be much appreciated. thanks

vadmeste:
The multipart upload was reading all parts and putting them
in memory before calling for thread manager to upload parts
in parallel.

This commit changes the behavior to read a part only when a
worker becomes ready to upload.

fixes #626

### DIFF
--- a/minio/thread_pool.py
+++ b/minio/thread_pool.py
@@ -42,19 +42,24 @@ class Worker(Thread):
         self.start()
 
     def run(self):
-        fast_quit = False
-        while not self.tasks_queue.empty():
-            func, args, kargs = self.tasks_queue.get()
-            if not fast_quit:
+        """ Continously receive tasks and execute them """
+        while True:
+            task = self.tasks_queue.get()
+            if task is None:
+                self.tasks_queue.task_done()
+                break
+            # No exception detected in any thread,
+            # continue the execution.
+            if self.exceptions_queue.empty():
                 try:
+                    # Execute the task
+                    func, args, kargs = task
                     result = func(*args, **kargs)
                     self.results_queue.put(result)
                 except Exception as e:
                     self.exceptions_queue.put(e)
-                    fast_quit = True
             # Mark this task as done, whether an exception happened or not
             self.tasks_queue.task_done()
-
 
 class ThreadPool:
     """ Pool of threads consuming tasks from a queue """
@@ -69,22 +74,21 @@ class ThreadPool:
         """ Add a task to the queue """
         self.tasks_queue.put((func, args, kargs))
 
-    def parallel_run(self, func, args_list):
-        """ Add a list of tasks to the queue """
-        for args in args_list:
-            self.add_task(func, args)
-
+    def start_parallel(self):
+        """ Prepare threads to run tasks"""
         for _ in range(self.num_threads):
             Worker(self.tasks_queue, self.results_queue, self.exceptions_queue)
 
+    def result(self):
+        """ Stop threads and return the result of all called tasks """
+        # Send None to all threads to cleanly stop them
+        for _ in range(self.num_threads):
+            self.tasks_queue.put(None)
         # Wait for completion of all the tasks in the queue
         self.tasks_queue.join()
         # Check if one of the thread raised an exception, if yes
         # raise it here in the function
         if not self.exceptions_queue.empty():
             raise self.exceptions_queue.get()
-
-    def result(self):
-        """ Return the result of all called tasks """
         return self.results_queue
 
--- a/minio/api.py
+++ b/minio/api.py
@@ -1528,7 +1528,7 @@ class Minio(object):
 
         # Instantiate a thread pool with 3 worker threads
         pool = ThreadPool(_PARALLEL_UPLOADERS)
-        parts_to_upload = []
+        pool.start_parallel()
 
         # Generate new parts and upload <= current_part_size until
         # part_number reaches total_parts_count calculated for the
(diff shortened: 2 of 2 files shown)


minio/thread_pool.py
```diff
@@ -42,11 +42,17 @@ def __init__(self, tasks_queue, results_queue, exceptions_queue):
         self.start()
 
     def run(self):
+        """ Continously receive tasks and execute them """
         fast_quit = False
-        while not self.tasks_queue.empty():
-            func, args, kargs = self.tasks_queue.get()
+        while True:
```
krisis: When a task throws an exception, `fast_quit` is set to `True`. IIUC, the loop proceeds to the next iteration, 'pops' the next task out of the queue __but__ exits the loop without executing it. So, the 'popped' task will __not__ be run by any of the workers in the thread pool? Is this behaviour acceptable?

minio/thread_pool.py
```diff
@@ -69,22 +74,21 @@ def add_task(self, func, *args, **kargs):
         """ Add a task to the queue """
         self.tasks_queue.put((func, args, kargs))
 
-    def parallel_run(self, func, args_list):
+    def start_parallel(self):
```
krisis: Now that we don't call add_task in this function, the comment needs to be changed accordingly.

minio/thread_pool.py
```diff
@@ -69,22 +74,21 @@ def add_task(self, func, *args, **kargs):
...
         """ Add a list of tasks to the queue """
-        for args in args_list:
-            self.add_task(func, args)
-
         for _ in range(self.num_threads):
             Worker(self.tasks_queue, self.results_queue, self.exceptions_queue)
 
+    def result(self):
+        """ Return the result of all called tasks """
+        # Send nil to all threads to cleanly stop them
```
krisis: (minor) nit: s/nil/None

minio/thread_pool.py
vadmeste: > When a task throws an exception, `fast_quit` is set to `True`. IIUC, the loop proceeds to the next iteration, 'pops' the next task out of the queue **but** exits the loop without executing it. So, the 'popped' task will **not** be run by any of the workers in the thread pool? Is this behaviour acceptable?

Yes, if one thread has an exception, it is not needed for other threads to continue uploads of parts

vadmeste: > Yes, if one thread has an exception, it is not needed for other threads to continue uploads of parts

I will modify the code a little bit, only the current thread is quitting, not other threads as I thought

vadmeste: > I will modify the code a little bit, only the current thread is quitting, not other threads as I thought

I removed fast_quit variable, now we will test directly on exceptions_queue and see if it is empty before continuing to run tasks

minio/thread_pool.py
vadmeste: > Now that we don't call add_task in this function, the comment needs to be changed accordingly.

Done

minio/thread_pool.py
vadmeste: > (minor) nit: s/nil/None

Done

minio/thread_pool.py
```diff
@@ -42,20 +42,25 @@ def __init__(self, tasks_queue, results_queue, exceptions_queue):
...
-            if not fast_quit:
+        """ Continously receive tasks and execute them """
+        while True:
+            task = self.tasks_queue.get()
+            if task is None:
+                self.tasks_queue.task_done()
+                break
+            # No exception detected in any thread,
+            # continue the execution.
+            if self.exceptions_queue.empty():
```
Praveenrajmani: I suspect a small contingency here, I doubt that the single worker is fed with all the tasks in the task queue here.
This condition `self.exceptions_queue.empty()` checks for the exception queue alone and allows all the tasks to actually reach `result = func(*args, **kargs)`. I can break out in points here
- Initially we have 3 worker threads up and ready by `pool.start_parallel()`
- Then we add the tasks one by one by `pool.add_task()`
- This condition here `self.exceptions_queue.empty()` allows any task to succeed and reach `result = func(*args, **kargs)`. Say, If there are 3 tasks and this condition here will pass until any exception is thrown yet. So any single worker will accept all the tasks. 
I think there should be something like `while not self.tasks_queue.empty()` so that the decision is made up on task_queue instead of exception_queue
 
I can be wrong or out of clarity. So please correct me if i'm wrong.

vadmeste: > * Say, If there are 3 tasks and this condition here will pass until any exception is thrown yet. So any single worker in the pool will accept all the tasks.

I didn't fully understand your point, let's talk privately

Praveenrajmani: Clear now .. Thanks Anis!

PR #713, opened 2018-10-26
Fix for the failing functional test cases for s3

### LINKED ISSUE #709: Functional test: minio-py while running in mint should exit on failure
Continuation of issue #708
If minio-py tests fail, 

1. Check if minio-py on failure returns 1 to the shell script that checks it's return value
2. If minio-py is not returning 1, we need to do so
3. If minio-py is returning 1, then we need to see why shell script is ignoring that

Praveenrajmani:
This PR fixes the following functional test cases which were failing on s3
- test_put_object
- test_stat_object

Taking into account that, metadata can be case-insensitive, the metadata is normalized for test cases so that it could work for both s3 and minio. `Minio` responds with canonicalized metadata, but s3 does not.

Fixes #709

### DIFF
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -564,6 +564,11 @@ def test_copy_object_unmodified_since(client, log_output):
     # Test passes
     print(log_output.json_report())
 
+def normalize_metadata(meta_data):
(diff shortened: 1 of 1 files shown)


vadmeste:
@Praveenrajmani you have a travis error which can go away if you rebase with master

PR #715, opened 2018-11-03
add progressbar

### LINKED ISSUE #601: I want see a progress bar

rkdls:
close #601

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -775,7 +775,7 @@ __Parameters__
 |``content_type``   |_string_ | Content type of the object. (optional, defaults to 'application/octet-stream').   |
 |``metadata``   |_dict_ | Any additional metadata. (optional, defaults to None). |
 |``sse`` |_dict_   |Server-Side Encryption headers (optional, defaults to None).   |
-|``progress`` |_bool_   |Display progress (optional, defaults to None).   |
+|``progress`` |_bool_   |Display progress (optional, defaults to False).   |
 
 __Return Value__
 
@@ -823,7 +823,7 @@ __Parameters__
 |``content_type``   |_string_ | Content type of the object (optional, defaults to 'application/octet-stream'). |
 |``metadata``   |_dict_ | Any additional metadata (optional, defaults to None). |
 |``sse`` |_dict_   |Server-Side Encryption headers (optional, defaults to None).   |
-|``progress`` |_bool_   |Display progress (optional, defaults to None).   |
+|``progress`` |_bool_   |Display progress (optional, defaults to False).   |
 
 __Return Value__
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -775,6 +775,7 @@ __Parameters__
 |``content_type``   |_string_ | Content type of the object. (optional, defaults to 'application/octet-stream').   |
 |``metadata``   |_dict_ | Any additional metadata. (optional, defaults to None). |
 |``sse`` |_dict_   |Server-Side Encryption headers (optional, defaults to None).   |
+|``progress`` |_bool_   |Display progress (optional, defaults to None).   |
 
 __Return Value__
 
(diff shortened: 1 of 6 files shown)


minio/progress.py
```diff
@@ -0,0 +1,96 @@
+import time
```
harshavardhana: You may add license header here.

minio/progress.py
```diff
@@ -0,0 +1,96 @@
...
+import sys
+from threading import Thread
+from .compat import queue, queue_empty
+
+
+class Progress(Thread):
```
harshavardhana: Each class should be documented properly since this is written like a generic progress printer.

minio/progress.py
```diff
@@ -0,0 +1,96 @@
...
+
+def format_meter(n, total, elapsed):
+    # n - number of finished iterations
+    # total - total number of iterations, or None
+    # elapsed - number of seconds passed since start
+    if n > total:
+        total = None
+
+    elapsed_str = format_interval(elapsed)
+    n_to_mb = n / 1024 / 1024
```
harshavardhana: Define all the constants, constants should be defined in the beginning of this file

minio/api.py
```diff
@@ -768,9 +771,13 @@ def put_object(self, bucket_name, object_name, data, length,
...
         metadata['Content-Type'] = 'application/octet-stream' if \
             not content_type else content_type
+        
+        if progress:
+            progress = Progress(total_size=length, file_name=object_name)
+
         if length > MIN_PART_SIZE:
             return self._stream_put_object(bucket_name, object_name,
-                                           data, length, metadata=metadata, sse=sse)
+                                           data, length, metadata=metadata, sse=sse, progress=progress)
```
harshavardhana: Overflowing the 80-100 column limit you should press enter and indent it within 80 column limit.

docs/API.md
```diff
@@ -775,6 +775,7 @@ __Parameters__
 |``content_type``   |_string_ | Content type of the object. (optional, defaults to 'application/octet-stream').   |
 |``metadata``   |_dict_ | Any additional metadata. (optional, defaults to None). |
 |``sse`` |_dict_   |Server-Side Encryption headers (optional, defaults to None).   |
+|``progress`` |_bool_   |Display progress (optional, defaults to None).   |
```
harshavardhana: If this is boolean, let's keep the default to False instead of None.

minio/progress.py
```diff
@@ -0,0 +1,166 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+minio.progress
+
+This module implements a progress while communicate minio server
```
harshavardhana: Better is `This module implements a progress printer while uploading objects`

minio/progress.py
```diff
@@ -0,0 +1,166 @@
...
+
+_REFRESH_CHAR = '\r'
+
+
+class Progress(Thread):
+    """
+        Constructs a :class:`Progress` object.
+
+        :param total_size: Total size of file.
+        :param file_name: Filename to be showed.
```
harshavardhana: call this perhaps `object_name` instead? since this is an object storage SDK?

vadmeste:
Thanks for your PR @rkdls 

Having a progress bar update is a great idea, however I would prefer to do not include the progress bar code in minio-py itself because users will then lose the capability to customize the progress bar itself for their own needs.

This is what I think we can do: instead of passing progress parameter as a boolean, we can let users to pass a function, let's call it update, so this latter can be called by minio-py to report data transfer update.

The progress code can be also committed in the documentation so new users can easily use it

What do you think @rkdls, @harshavardhana ?

rkdls:
yep. I know what you said, and I agree.

I will reopen new PR as soon as possible :)

PR #722, opened 2018-11-30
Metadata support for copy_object

### LINKED ISSUE #721: copyobject API doesn't support metadata
copyobject API doesn't support "metadata" arg which is part of S3 specifications. It is worth to mention that it is already supported in the other SDKs.

Praveenrajmani:
Fixes #721


sinhaashish:
Tested and LGTM.

PR #723, opened 2018-12-08
Add progress bar 

rkdls:
Changed the `put, fput` function to display the progress every 5MB.

Users can pass a custom function as an argument to 'progress' and display it on the screen.


minio/api.py
```diff
@@ -756,13 +758,19 @@ def put_object(self, bucket_name, object_name, data, length,
...
         """
 
         is_valid_sse_object(sse)
         is_valid_bucket_name(bucket_name)
         is_non_empty_string(object_name)
 
+        if progress:
+            if not isinstance(progress, Thread):
+                raise TypeError('Progress object should inherit the thread.')
+            progress.set_meta(total_length=length, object_name=object_name)
```
rkdls: Pass information to the progress object before sending the file.

minio/api.py
```diff
@@ -1491,21 +1500,23 @@ def _do_put_object(self, bucket_name, object_name, part_data,
             content_sha256=sha256_hex
         )
 
+        if progress:
+            progress.update(part_size)
```
rkdls: Update the 'progress' object here.

harshavardhana:
```suggestion
            # Update the 'progress' object with uploaded 'part_size'.
            progress.update(part_size)
```

minio/api.py
harshavardhana:
```suggestion
            # Set progress bar length and object name before upload
            progress.set_meta(total_length=length, object_name=object_name)
```

harshavardhana:
PTAL @vadmeste

PR #725, opened 2018-12-25
make_bucket: Use correct region when signing requests

### LINKED ISSUE #712: make_bucket: Specifying a non standard region fails with AWS S3
Running the code below will fail (notice the region parameter)

```
from minio import Minio

client = Minio('s3.amazonaws.com',
               access_key='YOUR-ACCESSKEYID',
               secret_key='YOUR-SECRETACCESSKEY')

# Make a new bucket
try:
    client.make_bucket('my-bucketname', 'us-west-1')
except ResponseError as err:
    print(err)
```

vadmeste:
make_bucket() uses new bucket location to sign make bucket requests,
however region should be used instead, and 'us-east-1' if no region
is specified.

Fixes #712

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -248,9 +248,11 @@ class Minio(object):
         """
         is_valid_bucket_name(bucket_name)
 
-        ## Region already set in constructor, validate if
-        ## caller requested bucket location is same.
+        # Default region for all requests.
+        region = 'us-east-1'
         if self._region:
+            region = self._region
+            # Validate if caller requested bucket location is same as current region
             if self._region != location:
                 raise InvalidArgumentError("Configured region {0}, requested"
                                            " {1}".format(self._region,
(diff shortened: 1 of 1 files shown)


minio/api.py
```diff
@@ -248,9 +248,11 @@ def make_bucket(self, bucket_name, location='us-east-1'):
...
         is_valid_bucket_name(bucket_name)
 
-        ## Region already set in constructor, validate if
-        ## caller requested bucket location is same.
+        # Default region for all requests.
+        region = 'us-east-1'
         if self._region:
+            region = self._region
+            # Validate if caller requested bucket location is same as current region
             if self._region != location:
```
kannappanr: Is this check needed?

vadmeste: > Is this check needed?

This comes from the old code but also it seems to me it is correct to do the check, because if the user explicitly sets a region when initializing Minio constructor, most likely all buckets that he needs to create will be in the same region.

PR #739, opened 2019-03-08
Fix #738 allow copy_object to set content-type

### LINKED ISSUE #738: allow `copy_object` with `content-type` in metadata
Trying to use `copy_object` to copy an object from one bucket to another.

To keep the target object well functional, I need to preserve `Content-Type` and `Content-Encoding` headers.

`Content-Encoding` is possible passing it's value via `metadata`.

However, I did not succeed with `Content-Type`.

The only solution was to edit sources of [minio/helpers.py](https://github.com/minio/minio-py/blob/master/minio/helpers.py#L667) function `is_supported_header` adding value "content-type".

```python
# returns true if a standard supported header
def is_supported_header(key):
    ## Supported headers for object.
    supported_headers = [
       "cache-control",
       "content-encoding",
       "content-type",  # added one
       "content-disposition",
       "content-language",
       "x-amz-website-redirect-location",
            ## Add more supported headers here.
        ]
    return key.lower() in supported_headers
```

This might be the solution, but I am not sure, it does not break anything else.

vlcinsky:
Fix issue #738

### DIFF
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -665,6 +665,7 @@ def is_supported_header(key):
     supported_headers = [
 	   "cache-control",
 	   "content-encoding",
+	   "content-type",
 	   "content-disposition",
 	   "content-language",
 	   "x-amz-website-redirect-location",
--- a/tests/unit/header_value_test.py
+++ b/tests/unit/header_value_test.py
@@ -127,8 +127,9 @@ class HeaderTests(TestCase):
                   'content-language':'fr'
                   }
         m = amzprefix_user_metadata(metadata)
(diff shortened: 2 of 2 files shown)


kannappanr:
@vlcinsky Please look at the solution suggested in #738 suggested by @poornas

kannappanr:
@vlcinsky Can you please fix Travis build issues?

vlcinsky:
Yes, I will fix it now.

Anyway, I will assume, the tests preventing usage of content-type in metadata are to be fixed. Hopefully this does not have other consequencies in other parts of the code.

PR #744, opened 2019-03-25
list-objects-v2: Add start_after parameter

### LINKED ISSUE #718: Add parameter to ListObjectsV2 to make the iteration interruptible
Hi guys

Currently the `ListObjectsV2` in SDK is not able to start a iteration in halfway, for some small cases it is ok. But in my project, a bucket have over 100k items, if any failing happened,  I have to re-start the whole iteration from the first item, it is really unacceptable for a large bucket

According to discussion of [issue 6825](https://github.com/minio/minio/issues/6825), go lang core API already had a `start-after` parameter, I think the SDK should support this too, for example:

``` python
objects = minioClient.list_objects_v2('mybucket', prefix='my-prefixname',
                              recursive=True)
for obj in objects:
    last_cursor = obj.cursor
```

if iteration failed, input cursor as `startAfter` will continue last task:

``` js
objects = minioClient.list_objects_v2('mybucket', prefix='my-prefixname',
                              recursive=True, start_cursor=last_cursor)
```

Looking forward for this feature, this is really important, thanks.

vadmeste:
When listing a bucket or a prefix which contains a lot of files,
it is not possible to pause/resume listing using list-objects-v2
API, adding start_after will help the developer listing from a
starting point without having to list everything again from the
beginning.

Fix #718

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -249,6 +249,7 @@ __Parameters__
 |``bucket_name`` | _string_ | Name of the bucket.|
 |``prefix``| _string_ |The prefix of the objects that should be listed.  Optional, default is None.|
 |``recursive``   | _bool_ |``True`` indicates recursive style listing and ``False`` indicates directory style listing delimited by '/'. Optional, default is False.|
+|``start_after``| _string_ | Starting listing after the specified path.  Optional, default is None.|
 
 __Return Value__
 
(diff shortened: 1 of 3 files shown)


vadmeste:
This is a breaking change!

vadmeste:
This not an API breaking anymore:

Old API:
    ```def list_objects_v2(self, bucket_name, prefix='', recursive=False)```
New API:	    
     ```def list_objects_v2(self, bucket_name, prefix='', recursive=False, start_after='')```

vadmeste:
ping @balamurugana @harshavardhana

PR #745, opened 2019-03-26
Add configurable part_size

harshavardhana:


minio/helpers.py
```diff
@@ -46,6 +46,7 @@
 # Constants
 MAX_MULTIPART_COUNT = 10000 # 10000 parts
 MAX_MULTIPART_OBJECT_SIZE = 5 * 1024 * 1024 * 1024 * 1024  # 5TiB
+MAX_PART_OBJECT_SIZE = 5 * 1024 * 1024 * 1024  # 5GiB
```
vadmeste: It is probably better to call this MAX_PART_SIZE, as MIN_PART_SIZE

minio/api.py
```diff
@@ -541,7 +542,8 @@ def listen_bucket_notification(self, bucket_name, prefix='', suffix='',
 
     def fput_object(self, bucket_name, object_name, file_path,
                     content_type='application/octet-stream',
-                    metadata=None, sse=None, progress=None):
+                    metadata=None, sse=None, progress=None,
+                    part_size=MIN_PART_SIZE):
```
vadmeste: I think it is better to add DEFAULT_PART_SIZE = MIN_PART_SIZE and use it instead of MIN_PART_SIZE, it is more clear and easy to change later.

harshavardhana: Done

minio/helpers.py
harshavardhana: Done

harshavardhana:
Updated the PR @vadmeste @Praveenrajmani

minio/api.py
```diff
@@ -795,6 +802,15 @@ def put_object(self, bucket_name, object_name, data, length,
         if length > MAX_MULTIPART_OBJECT_SIZE:
             raise InvalidArgumentError('Input content size is bigger '
                                        ' than allowed maximum of 5TiB.')
+
+        if part_size < DEFAULT_PART_SIZE:
```
vadmeste: This should be 'if part_size < MIN_PART_SIZE`, we cannot forbid a 5mb part if we set 64Mb as default part size in the future.

harshavardhana: That is the intention @vadmeste - why would want to go lower than DEFAULT there is no good reason.

harshavardhana: But i can keep for sanity sake.. MIN_PART_SIZE

minio/api.py
```diff
@@ -795,6 +802,15 @@ def put_object(self, bucket_name, object_name, data, length,
...
                                        ' than allowed maximum of 5TiB.')
+
+        if part_size < MIN_PART_SIZE:
+            raise InvalidArgumentError('Input part size is smaller '
+                                       ' than allowed minimum of 5MiB.')
+
+        if part_size > MAX_PART_SIZE:
+            raise InvalidArgumentError('Input part size is bigger '
+                                       ' than allowed maximum of 5GiB.')
+
```
kannappanr: Should there also be a check for `length` < `part_size*10000` ?

harshavardhana: Yes definitely @kannappanr

minio/api.py
```diff
@@ -795,6 +803,19 @@ def put_object(self, bucket_name, object_name, data, length,
         if length > MAX_MULTIPART_OBJECT_SIZE:
             raise InvalidArgumentError('Input content size is bigger '
                                        ' than allowed maximum of 5TiB.')
+
+        if (part_size * MAX_MULTIPART_COUNT) > length:
+            raise InvalidArgumentError('Part size '
```
kannappanr: @y4m4 comment is not complete. Shouldn't the check be `if length > (part_size * MAX_MULTIPART_COUNT):`

harshavardhana: no length is already checked @kannappanr

PR #747, opened 2019-04-03
Issue/7401

GvandeSteeg:
It was discussed in #7401 to add support for capitalisation inside minio buckets, as Minio server supports this now as well.

Example:
Previously accepted: `logfiles` 
Merging this would accept: `logfiles`, `LogFiles`, or any variation inbetween, as long as the _n_-th letter would be lowercase, i.e. `logfileS` would not be accepted.


GvandeSteeg:
Still not overly proficient with Git, it seems. I've only changed `helpers.py` and `minio_test.py`. But since I initially pulled from master that's probably why things are going bonkers

PR #749, opened 2019-04-06
Replaces Minio refs with MinIO and minio.io links with min.io

ebozduman:
Official company name `Minio` became `MinIO`. 
This change requires all references in the code and documentation to be modified as `MinIO` and the links, `....://xxx.minio.io`, to be `....://xxx.min.io`.

Information about the change:
Used Visual Studio Code Editor regular expressions to find and replace those patterns that match the text in all minio-py repo files. 

Here are the two regular expressions used to complete the job in 2 steps:

First, choose "Find in Files' from the Edit menu and activate `Match Case` and `Use Regular Expression` options.
1.  Replace the links using the following regular expressions:

| Find | Replace |
| ------ | ----------- |
| \bminio.io\b | min.io |

The above regular expression replaces the links that start with either `slack` or `docs` or ``play` or `dl` or `blog` or nothing before `minio.io`.

2. 
| Find | Replace |
| ------ | ----------- |
| \bMinio\b(?!(\('\|\([eA-Z1])) | MinIO |
| import MinIO | import Minio |

1st regexp above touches `from minio import Minio`, so we need to put `from minio import MinIO` back to what it was. That's why we have the 2nd regexp.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -49,7 +49,7 @@ EOF
 ```
 
 #### Sign
-Sign the release artifacts, this step requires you to have access to Minio's trusted private key.
+Sign the release artifacts, this step requires you to have access to MinIO's trusted private key.
 ```sh
 $ export GNUPGHOME=/media/${USER}/minio/trusted
 $ gpg --detach-sign -a dist/minio-2.2.5.tar.gz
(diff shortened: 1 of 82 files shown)


donatello:
Broke the build.

kannappanr:
Yes, api.py needs to change

PR #750, opened 2019-04-08
Replaces Minio refs with MinIO and minio.io links with min.io

kannappanr:
Replace minio.io with min.io and Minio with MinIO

PR is based on https://github.com/minio/minio-py/pull/749

### DIFF
--- a/examples/progress.py
+++ b/examples/progress.py
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 """
-This module implements a progress printer while communicate minio server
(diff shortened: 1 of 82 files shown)


examples/progress.py
```diff
@@ -17,7 +17,7 @@
 """
 This module implements a progress printer while communicate minio server
```
Praveenrajmani: minio -> MinIO

kannappanr: done

PR #769, opened 2019-06-10
Add test case for Object name validation

sinhaashish:
Added ad test case to check if the object name has prefix as a slash.


kannappanr:
Closing this PR as https://github.com/minio/minio/pull/7746 doesn't need this test anymore

PR #777, opened 2019-07-12
Add support for ISO time with and without fractional seconds.

mikijov:
Some non Amazon S3 implementations are not consistent about including fractional seconds in the timestamp. The change allows parsing of both `2019-07-10T14:23:39.231Z` and `2019-07-10T14:23:39Z`.

### DIFF
--- a/minio/parsers.py
+++ b/minio/parsers.py
@@ -360,6 +360,10 @@ def _iso8601_to_localized_time(date_string):
     :param date_string: iso8601 formatted date string.
     :return: :class:`datetime.datetime`
     """
+
+    # Handle timestamps with and without fractional seconds. Some non-AWS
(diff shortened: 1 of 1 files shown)


harshavardhana:
Can you state which Amazon S3 implementation is this ? Please mention them in the comment. We may way to remove this code in future, as this should be seen as a workaround.

mikijov:
Added comment into the code. I encountered this issue on Dell EMC ECS.

mikijov:
Build job failed due to some timeout, I have to close/re-open the PR to trigger the build.

mikijov:
Hi @ebozduman, @vadmeste, are there any concerns merging this? We actually need this in production, and we are struggling patching minio-py on the fly...

mikijov:
Thanks for approval. Will someone be able to merge it?

PR #784, opened 2019-08-08
api access data removed and default values ​​set

felipecolen:


examples/listen_notification.py
```diff
@@ -20,8 +20,8 @@
 from minio import Minio
 
 client = Minio('play.min.io',
-               access_key='Q3AM3UQ867SPQQA43P2F',
-               secret_key='zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG')
+               access_key='YOUR-ACCESSKEYID',
```
harshavardhana: I don't think this change is needed in this PR. play.min.io credentials are set for a reason here.

harshavardhana:
Would you be able to address the comment here @felipecolen ? - if not shall we close this PR.

PR #799, opened 2019-09-15
Fix: Missing space between words in error message

innerlee:
Improve error message.

current problem:

image

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -663,13 +663,13 @@ class Minio(object):
 
             # Verify if we wrote data properly.
             if total_written < content_size:
-                msg = 'Data written {0} bytes is smaller than the' \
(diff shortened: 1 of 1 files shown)


kannappanr:
Can the title and commit message be changed to `Fix: Missing space between words in error message` ? That seems more appropriate here.

innerlee:
For the commit message, maybe the maintainer can use [squash pr](https://github.blog/2016-04-01-squash-your-commits/) to change the message to appropriate ones.

Or, I can create a new pr by editing the source file again.

I'm trying to avoid cloning the repo to local just for a force push.

harshavardhana:
We know how to fix the title but we like original works with clarity when submitted as PR. A PR wording is as important as the code written in the PR, it should reflect the original work as much as possible.

In this case however we can fix the title ourselves.

PR #802, opened 2019-09-26
Strip port 80/443 from host

edward-codecov:
Supplying port 80 or 443 with host (ex. `storage.googleapis.com:443`) fails in v4 signing. Remove these default ports to prevent signature matching failure.

Similar issue:
https://github.com/aws/aws-cli/issues/2883

### DIFF
--- a/minio/signer.py
+++ b/minio/signer.py
@@ -357,7 +357,13 @@ def generate_authorization_header(access_key, date, region,
     return ' '.join(auth_header)
 
 def remove_default_port(parsed_url):
-    if parsed_url.port is 80 or 443:
+    default_ports = {
+        'http': 80,
+        'https': 443
+    }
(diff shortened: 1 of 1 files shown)


sinhaashish:
@edward-codecov  Thanks for you PR.
IMO you can try something like 
```
 url_parts = urlsplit(url)
        default_ports = {
            'http': 80,
            'https': 443
        }
        if any(url_parts.scheme == scheme and url_parts.port == port
               for scheme, port in default_ports.items()):
            # No need to include the port if it's the default port.
            return url_parts.hostname
```
This way we keep a track that if port 80 is stripped then it belongs to http and 443 for https. 
This is a better way than just stripping with the use of if condition like the one you did  `if parsed_url.port is 80 or 443:`

Issue #810, opened 2019-10-21
minio.error.SignatureDoesNotMatch: SignatureDoesNotMatch: message: The request signature we calculated does not match the signature you provided. when upload to ceph radosgw


phchoi:
Not sure if the issue I faced is the same as this #724 , it throws the exception when a file larger than 5MB is being uploaded to ceph radosgw by using minio python API but not via mc client

Environment:

python 2.7.5,
minio python library version 5.0.1
minio client: minio version RELEASE.2019-10-09T22-54-57Z
I followed the same troubleshoot methods mentioned in #724 (comment) and had the log attached ([minio-py.log](https://github.com/minio/minio-py/files/3750175/minio-py.log)). I also attached the same debug log coming from mc client ([mc.log](https://github.com/minio/minio-py/files/3750161/mc.log))

Any hints or help will be much appreciated, thanks a lots.

kannappanr:
@phchoi thanks for testing with `5.0.1`. We will take a look.

phchoi:
Thanks for the quick turnaround, @kannappanr

aliir74:
I also have this problem with uploading large file into ceph

vadmeste:
> I also have this problem with uploading large file into ceph

Yes, we have a PR which fixes the issue under the review

phchoi:
Thanks @vadmeste  !

aliir74:
Thanks for fast fix :)
How can I use this fix with `pip install minio`?

PR #811, opened 2019-10-21
signV4: Avoid encoding ~ when found in the query part of the url

vadmeste:
Ceph radosgw seems to include ~ in the upload ID when the client
initiates a new multipart upload.

minio-py was wrongly encoding ~ in the query part of the url.

This PR ensures that to decode %7E to ~ before signing the request.

Fixes https://github.com/minio/minio-py/issues/810

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -331,7 +331,8 @@ class Minio(object):
                           headers, self._access_key,
                           self._secret_key,
                           self._session_token,
-                          content_sha256_hex)
+                          content_sha256_hex,
+                          datetime.utcnow())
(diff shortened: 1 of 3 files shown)


sinhaashish:
@vadmeste  Can you check the failure.

vadmeste:
@Praveenrajmani @sinhaashish PTAL

harshavardhana:
Can you add a unit test for this @vadmeste ?

vadmeste:
> Can you add a unit test for this @vadmeste ?

Done

PR #812, opened 2019-10-22
tests: More fixes for better support of Github CI

vadmeste:
1. Explicity create and pass temporary directories in tests

2. Download Minio binary for both Linux & Windows and avoid
   testing directly against Play server


harshavardhana:
Looks like functional tests hang and GitHub cancels the CI

```
{"function": "stat_object(bucket_name, object_name, sse)", "status": "PASS", "duration": 81673, "args": {"object_name": "2bdc7324-fdd7-45a9-b10c-cfe5d56fb30a", "length": 11534336, "bucket_name": "minio-py-test-0d54bb00-0e07-478e-b49d-47a6170caf70", "content_type": "application/octet-stream", "data": "LimitedRandomReader(MB_11)", "metadata": {"X-Amz-Meta-Testing": "value"}}, "name": "minio-py:test_stat_object_with_SSE-C"}
##[error]The operation was canceled.
```

vadmeste:
> Looks like functional tests hang and GitHub cancels the CI

It looks like Github cancels when there is one failed job, that's why it has two netural checks

vadmeste:
jobs.<job_id>.strategy.fail-fast
When set to true, GitHub cancels all in-progress jobs if any matrix job fails. Default: true

https://help.github.com/en/github/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions

vadmeste:
I made some changes but still struggling with connection reset error..

This is the log in Play server

```
 API: PutObject(bucket=minio-py-test-67de652e-b8e1-4608-8187-44767afc0c30, object=efe9385f-10bd-4747-a252-29802e82fd56)
 Time: 16:04:03 UTC 10/23/2019
 DeploymentID: 1d356152-c02f-4e2d-8f55-8df8f5c4ed67
 RequestID: 15D0522D350AAA3E
 RemoteHost: 197.14.48.119
 UserAgent: MinIO (Linux; x86_64) minio-py/5.0.2
 Error: #has incomplete body
        5: cmd/xl-v1-object.go:639:cmd.xlObjects.putObject()
        4: cmd/xl-v1-object.go:514:cmd.xlObjects.PutObject()
        3: cmd/xl-sets.go:653:cmd.(*xlSets).PutObject()
        2: cmd/object-handlers.go:1257:cmd.objectAPIHandlers.PutObjectHandler()
        1: net/http/server.go:2007:http.HandlerFunc.ServeHTTP()


```

harshavardhana:
I think you should download MinIO locally and run it like minio-go 

https://github.com/minio/minio-go/blob/master/.travis.yml#L19

vadmeste:
> I think you should download MinIO locally and run it like minio-go
> 
> https://github.com/minio/minio-go/blob/master/.travis.yml#L19

yes, good idea

vadmeste:
This is finally ready to review & merge.

.github/workflows/pythonpackage-windows.yml
```diff
@@ -0,0 +1,49 @@
...
+        pyflakes minio/*.py
+        python setup.py install
+        python setup.py nosetests
+    - name: Test with functional tests
+      env:
+        MINT_MODE: full
+        SERVER_ENDPOINT: localhost:9000
+        ACCESS_KEY: minio
+        SECRET_KEY: minio123
+        ENABLE_HTTPS: 0
```
harshavardhana: It is better to keep a cert and enable HTTPS for encryption-related tests @vadmeste just like ubuntu-latest

.github/workflows/pythonpackage-linux.yml
```diff
@@ -36,4 +40,15 @@ jobs:
...
+        SERVER_ENDPOINT: localhost:9000
+        ACCESS_KEY: minio
+        SECRET_KEY: minio123
+        ENABLE_HTTPS: 1
+        MINIO_ACCESS_KEY: minio
+        MINIO_SECRET_KEY: minio123
+      run: |
+          wget -O $HOME/minio https://dl.min.io/server/minio/release/linux-amd64/minio
+          chmod +x $HOME/minio
+          $HOME/minio server /tmp/xl/{1..4} &
```
harshavardhana: erasure coding is perhaps an overkill just use `fs` mode fine? @vadmeste

.github/workflows/pythonpackage-linux.yml
```diff
@@ -36,4 +40,15 @@ jobs:
...
+          export SSL_CERT_FILE=$HOME/.minio/certs/public.crt
```
harshavardhana: set in the `env:` section

.github/workflows/pythonpackage-linux.yml
```diff
@@ -28,6 +28,10 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install urllib3 certifi pytz pyflakes faker nose
+    - name: Generate TLS certificate
+      run: |
+        mkdir -p $HOME/.minio/certs/
```
harshavardhana: This is auto-created I am not sure we need to create this. Also lets use a non-homedir such that we don't need to rely on homedir location perhaps pass `--certs-dir` to MinIO ?

.github/workflows/pythonpackage-windows.yml
vadmeste: > It is better to keep a cert and enable HTTPS for encryption-related tests @vadmeste just like ubuntu-latest

It is little hard to generate certificates in Windows, it seems it is easier to create a certificate and commit in minio-py, like what's done in minio-go

harshavardhana:
why generate @vadmeste commit to the minio-py since we have to use a static 127.0.0.1 IP anyways

.github/workflows/pythonpackage-windows.yml
harshavardhana: Commit to the minio-py don't autogenerate it @vadmeste

harshavardhana:
NOTE: on windows you need it to be 127.0.0.1 not localhost @vadmeste so cert should be IP SANs

vadmeste:
> NOTE: on windows you need it to be 127.0.0.1 not localhost @vadmeste so cert should be IP SANs

It looks like it works fine with localhost

PR #819, opened 2019-10-31
Do not need to set max-keys, max-uploads and max-parts

harshavardhana:
MinIO server supports now 10k elements in the list output,
there is no reason for client to limit itself let the
default set of keys be returned and limited by server
implementation

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -924,7 +924,6 @@ class Minio(object):
 
         # Initialize query parameters.
         query = {
-            'max-keys': '1000',
             'prefix': prefix
         }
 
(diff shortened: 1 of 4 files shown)


kannappanr:
```
$ mc admin trace myminio
^[[A12:23:32.990 [200 OK] s3.GetBucketLocation localhost:9000/test?location=  127.0.0.1         526µs       ↑ 82 B ↓ 382 B
12:23:32.995 [200 OK] s3.ListObjectsV1 localhost:9000/test/?max-keys=1000&prefix=  127.0.0.1         511µs       ↑ 93 B ↓ 507 B
12:23:32.998 [200 OK] s3.ListObjectsV2 localhost:9000/test/?list-type=2&prefix=&start-after=  127.0.0.1         409µs       ↑ 93 B ↓ 513 B

```
Listing in listobjectsV1 on the client side is still listing only 1000 objects at a time?

harshavardhana:
@kannappanr that's not possible unless you are using an old version.

PR #821, opened 2019-11-08
Increase presigned timeouts for functional tests

harshavardhana:
Fixes https://github.com/minio/minio/issues/8092

### DIFF
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -1307,7 +1307,7 @@ def test_presigned_get_object_default_expiry(client, log_output):
     print(log_output.json_report())
 
 
-def test_presigned_get_object_expiry_5sec(client, log_output):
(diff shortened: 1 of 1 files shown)


tests/functional/tests.py
```diff
@@ -1327,7 +1327,7 @@ def test_presigned_get_object_expiry_5sec(client, log_output):
 
         presigned_get_object_url = client.presigned_get_object(bucket_name,
                                                                object_name,
-                                                               timedelta(seconds=5))
+                                                               timedelta(seconds=120))
```
kannappanr: Function names here says `test_presigned_get_object_expiry_5sec`, that should be changed. Other functions where we have changed the code should also have their names changed to reflect the changes made.

harshavardhana: Done

harshavardhana: Done

PR #827, opened 2019-11-27
Add support for Amazon S3 Transfer Acceleration for presigned urls

cosmosgenius:
This PR add support for Amazon S3 Transfer Acceleration.

Acceleration is enable by using `client.use_s3_accelerate(True)`. This function is a noop is the `endpoint_url` is not point to `s3.amazonaws.com`.

When presigned generators are called eg. `client.presigned_get_object` internally i am checking the following.
- if S3 Accelerate is enable using `self._enable_s3_accelerate`
- if the bucket has accelerate configured using `self.is_bucket_acclerate_configured`
If both the conditions are satisfied i switch the endpoint_url (pass to `get_target_url`) to `://s3-accelerate.amazonaws.com`

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -163,8 +163,10 @@ class Minio(object):
 
         # Default is a secured connection.
         endpoint_url = 'https://' + endpoint
+        accelerate_endpoint_url = 'https://s3-accelerate.amazonaws.com'
         if not secure:
             endpoint_url = 'http://' + endpoint
+            accelerate_endpoint_url = 'http://s3-accelerate.amazonaws.com'
 
         # Parse url endpoints.
         url_components = urlsplit(endpoint_url)
@@ -182,6 +184,8 @@ class Minio(object):
         self._session_token = session_token
         self._user_agent = _DEFAULT_USER_AGENT
         self._trace_output_stream = None
+        self._enable_s3_accelerate = False
+        self._accelerate_endpoint_url = accelerate_endpoint_url
 
         # Set credentials if possible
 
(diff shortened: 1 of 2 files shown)


cosmosgenius:
I want to write tests but i'm bit confused on where to start.

cosmosgenius:
Hey is there anything else required in this PR? Been using a patched forked version for my usecase for a while now, but would like to switch to the official package.

minio/api.py
```diff
@@ -1355,7 +1396,12 @@ def presigned_url(self, method,
                                        ' {0} secs'.format(_MAX_EXPIRY_TIME))
 
         region = self._get_bucket_region(bucket_name)
-        url = get_target_url(self._endpoint_url,
+        endpoint_url = self._endpoint_url
+        if self._enable_s3_accelerate:
+            if self.is_bucket_accelerate_configured(bucket_name):
```
harshavardhana: This type of code is expensive when transfer acceleration is not enabled.  You don't need to implement the verification for parser acceleration and let the server fail.

cosmosgenius: So I am specifically checking this because acceleration is enabled at global level and not per bucket. 

If a bucket is not configured to use acceleration it uses the normal url. Eg. some internal bucket ("bucket A") where acceleration is disabled but for user uploads ("bucket B") acceleration is configured. Using the Minio i want to access both the buckets, but use acceleration when possible, if it is enabled in Minio.

harshavardhana: Let it fail on the server side.  I don't think we should have such type of checks.

cosmosgenius: Hey, I have removed the acceleration check

PR #833, opened 2019-12-15
Put object without length

chickeaterbanana:
put_object without length parameter. If you want to put a file-like-object and do not know the size before running this method it would be useful to have this method.


Praveenrajmani:
@chickeaterbanana Can you revert the linting here? Since you already have a PR for that..

Praveenrajmani:
Ping @chickeaterbanana

minio/helpers.py
```diff
@@ -187,6 +187,21 @@ def read_full(data, size):
...
+
+def read_complete_file_in_chunks(data, size):
+    """
+    read complete file in chunks
+    """
+    content = read_full(data, size)
+    read_size = len(content)
+    yield content
+
+    while read_size >= size:
```
Praveenrajmani: Can we have `while read_size == size` instead? `read_full` exactly reads `size` bytes from the reader

minio/api.py
```diff
@@ -860,14 +872,28 @@ def put_object(self, bucket_name, object_name, data, length,
...
-            return self._stream_put_object(bucket_name, object_name,
-                                           data, length, metadata=metadata,
-                                           sse=sse, progress=progress,
-                                           part_size=part_size)
+        current_data = []
+        iterable = iter(read_complete_file_in_chunks(data, part_size))
+        stream = False
+        try:
+            current_data.append(next(iterable))
+            stream = True
```
Praveenrajmani: shouldn't the stream be `False` if the `len(current_data) == 1`

chickeaterbanana: current_data has alway a size of 1. I used a list to use `itertools.chain` later(line 888). An alternative woul be:
```
           current_data = None
           iterable = iter(read_complete_file_in_chunks(data, part_size))
           stream = False
           try:
               current_data = next(iterable)
               stream = True
           except StopIteration:
               pass # we catch the StopIteration to know if there are more data than part_size or not
           
           ...
           
           itertools.chain([current_data], iterable), # line 888 we want to chain the already read data with the rest of the chunks
           ...
```

minio/helpers.py
chickeaterbanana: That's possible. I'll change it this evening(Europe/Berlin).

minio/api.py
Praveenrajmani: @chickeaterbanana what i meant was - if `iterable` yields only one element (for a smaller object - single put), `stream = True` will be true always and we would try multipart upload for those objects.

And stream will be set to False only when next(iterable) throws `StopIteration` for the first iteration itself

harshavardhana:
Let's not support this scenario, it's not useful in case of minio-py as it may lead to allocations and overwhelm the system.

It is okay to support in languages like Java and Go where we have a tiny bit more control on the memory. AWS S3 doesn't support Put without content length, monkey patching it for client side is not useful for minio-py 's usecases.

minio/api.py
Praveenrajmani: adding `current_data.append(next(iterable))` after `stream = False` would solve this i supppose. Meaning, if second iteration of next(iterable) did not throw `StopIteration`, i can be identified as object stream (stream = True)

chickeaterbanana: You are correct. To check this I have added a unittest. I corrected it.

chickeaterbanana:
@harshavardhana I'm not quite sure why this will lead to allocations? If I'm right this code is allocating max 2*part_size at the beginning of the upload, afterwards its only allocating part_size data. So the code is adding the overhead of one time part_size. Or do I have missed something?!

harshavardhana:
@chickeaterbanana uploading an object without content length would require allocating 512MiB to support 10,000 parts limit. With parallel workers this becomes easily 512 * n workers 

This is too much for python to handle, if you haven't implemented it this way then I think the PR may be incorrect to begin with. All in all its overkill to support put without content length when underlying API doesn't support it.

chickeaterbanana:
The underlying API supports it. You must not say how many parts your multipart file will have. So you don't have to know how many parts your file will have. I haven't changed something on the base logic how many parts in parallel will be uploaded. But the logic how many parts in parallel you want to upload isn't related to the size oft the file. Your example assumes you want so many worker that the whole file is uploaded in parallel with as many worker as file has parts. That is a bad idea in all programming languages because you have to hold the object to upload in memory you want to upload so end up to have the whole file in memory. Another reason not to do that are spinning disks, if you find a way not to store the part you are uploading in memory you have a lot of random-io on disk a sequential read is a better idea. The third reason is that the upload process is more or less io blocking and not CPU so more worker can end in a total overloaded network interface and it can happen that some of your tcp-connection will timeout or in other ways stop working.

harshavardhana:
I mean S3 API doesn't support it. So all this uplifting on client side is not worth it. You can set the right length from the get go and nothing will change. 

I would prefer a full stream APi but unfortunately S3 API doesn't have it.

chickeaterbanana:
There are some use cases where this can be useful for example a command line interface. You want to do something like that:
`cat /var/log/vey_big_file | grep "some_filter" | minio-py-cli upload server bucket object -`
or something like that:
`curl -s https://server-with-stream-file.io | minio-py-cli upload server bucket object -`

You don't know how big this file will be, if you haven't the space or a very slow disk you don't want to save the file in a temporary file on disk.

All stream apis I have seen so far, have enforced to stream the data in chunks. Thats the same interface S3 supports, they allow to stream data with a minimum_chunk_size of 5MB. And thats clever from AWS because a lot of requests are better to scale than a lot of streams with open connections. 

Another benefit of chunks is that you can buffer the chunk and read/write the optimal chunk_size of data to your disk. If you think big in TB for example, 5MB is a very small chunk_size. You are right if you think in small files. But if you have a lot of small files you don't need a stream/chunked-interface.

Some links related to this thema:
https://en.wikipedia.org/wiki/C10k_problem

harshavardhana:
All those do not make sense for this SDK. We already have a tool which does this `mc` you are better off using it. Complicating this library is not our goal, so it doesn't really what those minor use cases are they do not warrant this change.

harshavardhana:
While we appreciate that you spent time on writing this PR, it would have beneficial to ask us on our https://slack.min.io on our intentions of keeping our SDKs simple, as I stated before minio-go and minio-java handle these cases already but we did that with good conscience, it doesn't need to be replicated into minio-py

If you want to still use it, fork this library and keep the patch yourself. We don't see a need to address these niche usecases.

harshavardhana:
Closing this PR as out of scope for minio-py

PR #836, opened 2019-12-23
Fix detecting a valid endpoint with new Python versions

### LINKED ISSUE #835: is_valid_endpoint(endpoint)
Just want to clarify with the function as shown in title.
Python version: 3.7.6
Minio version: 5.0.5

ipaddr = '43.74.4.35:9000', my server ip address

mio = Minio(ipaddr, access_key=id,secret_key=pass,secure=False)
mio.fput_object(bucketname,filename,pathtofile)

returns error InvalidEndpointError:message: Hostname cannot have a scheme.

I did some digging and when send through the is_valid_endpoint(endpoint) function it will return:-
scheme: urlsplit(endpoint)[0] is 43.74.4.35
netloc: urlsplit(endpoint)[1] is empty
path: urlsplit(endpoint)[2] is 9000
query: urlsplit(endpoint)[3] is empty
fragment: urlsplit(endpoint)[4] is empty

I think the urlsplit is recognizing the ip address as scheme because it is before the ':'

Is it a bug or it is just my environment?
I just pip install minio.

EDIT:

i remove the port from ip address
ipaddr = '43.74.4.35'
but it return error
ResponseError: code: , message: , bucket_name: broadcast, object_name: None, request_id: , host_id: , region:

vadmeste:
Fixes #835 

In newer versions of Python, such as 3.7.6, the standard urlparse does not parse endpoint 'media:9004' as it is used to be.

```
Python 3.7.6 (default, Dec 20 2019, 22:50:33) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import urllib.parse
>>> urlsplit = urllib.parse.urlsplit
>>> urlsplit('media:9000')
SplitResult(scheme='media', netloc='', path='9000', query='', fragment='')
```

The new Python documentation states that we need to add '//' to enforce parsing media as a hostname, which is done in this PR.

### DIFF
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -302,7 +302,7 @@ def is_valid_endpoint(endpoint):
         if u.scheme:
             raise InvalidEndpointError('Hostname cannot have a scheme.')
 
-        hostname = u.netloc.split(':')[0]
+        hostname = u.hostname
         if hostname is None:
             raise InvalidEndpointError('Hostname cannot be empty.')
 
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -288,10 +288,21 @@ def is_valid_endpoint(endpoint):
        otherwise.
     """
     try:
-        if urlsplit(endpoint).scheme:
+        if not '//' in endpoint:
+            # Having '//' in the beginning of the endpoint enforce
+            # urlsplit to consider the endpoint as a netloc according
+            # to this quote in docs.python.org/3/library/urllib.parse.html:
+            #    Following the syntax specifications in RFC 1808, urlparse
+            #    recognizes a netloc only if it is properly introduced by ‘//’.
+            #    Otherwise the input is presumed to be a relative URL and thus
+            #    to start with a path component.
+            endpoint = '//' + endpoint
+
+        u = urlsplit(endpoint)
+        if u.scheme:
             raise InvalidEndpointError('Hostname cannot have a scheme.')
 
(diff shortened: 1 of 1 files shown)


minio/helpers.py
```diff
@@ -288,10 +288,21 @@ def is_valid_endpoint(endpoint):
...
+            #    Otherwise the input is presumed to be a relative URL and thus
+            #    to start with a path component.
+            endpoint = '//' + endpoint
+
+        u = urlsplit(endpoint)
+        if u.scheme:
             raise InvalidEndpointError('Hostname cannot have a scheme.')
 
-        hostname = endpoint.split(':')[0]
+        hostname = u.netloc.split(':')[0]
```
wtayyeb: use `hostname = u.hostname` as it is more standard way to get the hostname,
general netloc is like below but not limited to:
- host:port
- user:pass@host:port
- IPv6  [2001:db8::8a2e:370:7334]:port

as you see for ipv6 your logic will fail.

vadmeste: Thanks for the tip @wtayyeb

vadmeste:
This PR is not WIP anymore

sergii-tsymbal-exa:
Thanks for the fix. Could we initiate a new minio-py release (5.0.6) on pypi containing changes, as this issue is a blocker for people using python 3.7.6 or 3.8.1

PR #840, opened 2020-01-08
fix: relax bucket_name validation for existing buckets

### LINKED ISSUE #838: Create option for ignoring S3 naming rules
Minio supports exposing existing datasets on NAS (for the NAS gateway function).  The data that's already there, especially if it was created in a Windows environment, but certainly possible in a Linux environment, likely isn't going to conform to S3 standards- meaning it will likely have capital letters, or symbols that don't adhere to the standard. 

It would be MOST useful if there was a "relaxed" setting that allowed more, if not all, POSIX-compliant file and directory names to be used, specifically for API calls. 

I.E. My NAS has a directory of testing data called "WindowsUpdateLogs".  That's not compliant.  But when I start Minio against that directory, the browser shows it fine and will serve it up fine. 

Performing a list_buckets() from Python shows the bucket with the "invalid" name.  But if I try to list the objects in that bucket, it errors out as the bucket name I've passed is... invalid.  So that's a problem. 

Any ideas? or am I right to be requesting this as an RFE?

ebozduman:
Fixes #838.

To fix this issue, the implementation difference between minio-go and minio-py is needed to be addressed. 
With the fix, just like in minio-go, dev-engineers are now able to call strict or looser bucket name validation depending on the api.
Strict validation is only used for `make_bucket` api, and looser bucket name validation is used for the rest.
The relaxation accomplished with the new approach is listed below:

- Capital letters of English alphabet
- underscore character, `_`
- colon character, `:`

Now, they are also valid characters.

### DIFF
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -55,9 +55,15 @@ MAX_POOL_SIZE = 10
 MIN_PART_SIZE = 5 * 1024 * 1024  # 5MiB
 DEFAULT_PART_SIZE = MIN_PART_SIZE  # Currently its 5MiB
 
-_VALID_BUCKETNAME_REGEX = re.compile('^[a-z0-9][a-z0-9\\.\\-]+[a-z0-9]$')
+_VALID_BUCKETNAME_REGEX = re.compile(
+    '^[A-Za-z0-9][A-Za-z0-9\\.\\-\\_\\:]{1,61}[A-Za-z0-9]$')
+_VALID_BUCKETNAME_STRICT_REGEX = re.compile(
+    '^[a-z0-9][a-z0-9\\.\\-]{1,61}[a-z0-9]$')
+_VALID_IP_ADDRESS = re.compile(
+    '^(\d+\.){3}\d+$')
 _ALLOWED_HOSTNAME_REGEX = re.compile(
-    '^((?!-)(?!_)[A-Z_\\d-]{1,63}(?<!-)(?<!_)\\.)*((?!_)(?!-)[A-Z_\\d-]{1,63}(?<!-)(?<!_))$',
+    '^((?!-)(?!_)[A-Z_\\d-]{1,63}(?<!-)(?<!_)\\.)*((?!_)(?!-)' +
+    '[A-Z_\\d-]{1,63}(?<!-)(?<!_))$',
     re.IGNORECASE)
 
 _EXTRACT_REGION_REGEX = re.compile('s3[.-]?(.+?).amazonaws.com')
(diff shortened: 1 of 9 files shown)


tests/unit/minio_test.py
```diff
@@ -28,23 +28,23 @@
 class ValidBucketName(TestCase):
     @raises(InvalidBucketError)
     def test_bucket_name(self):
-        is_valid_bucket_name('bucketName')
+        is_valid_bucket_name('bucketName', False)
```
vadmeste: @ebozduman, unit tests won't pass here, you need to pass True instead of False in your change in this file

vadmeste: (Run this `bash tests/unit_test.sh`)

minio/helpers.py
```diff
@@ -351,16 +357,33 @@ def is_valid_bucket_name(bucket_name):
...
-                                 ' periods.')
+
+    match = _VALID_IP_ADDRESS.match(bucket_name)
+    if match:
+        raise InvalidBucketError('Bucket name cannot be an ip address')
+
+    unallowed_successive_chars = ['..', '.-', '-.']
+    if any(x in bucket_name for x in unallowed_successive_chars):
+        raise InvalidBucketError('Bucket name contains invalid '
+                                 'successive chars ("..", ".-", "-.").')
```
vadmeste: More easier to do this:
`'success chars ' + str(unallowed_successive_chars)`

harshavardhana:
@ebozduman the PR title needs to describes what it's fixing.

minio/helpers.py
```diff
@@ -351,16 +357,33 @@ def is_valid_bucket_name(bucket_name):
...
+
+    match = _VALID_BUCKETNAME_STRICT_REGEX.match(bucket_name)
```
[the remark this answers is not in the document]
ebozduman: Good point. This line was supposed to be inside the `strict` check.
Done.

tests/unit/minio_test.py
ebozduman: The fix is to introduce a relaxed validation capability for all apis, except `make_bucket` api. So, instead of changing to strict validation, I changed the tests.

minio/helpers.py
ebozduman: Done.

ebozduman:
I've reproduced the problem with `list_objects` api. It was previously considering a bucket name with all upper case letters as invalid.
With the fix, the same bucket name with all upper case letters is considered to be valid.

Again, the new validation approach relaxes the following restrictions:
- Capital letters of English alphabet are added to the valid characters list
- underscore character, `_`, is added to the valid characters list
- colon character, `:`, is added to the valid characters list

harshavardhana:
@ebozduman please fix the title, you should indicate in a more concise manner such as "fix: relax bucket_name validation for existing buckets" is better than "matches minio-go implementation"

kannappanr:
ping @vadmeste

PR #855, opened 2020-02-14
Support to put/get/delete default encryption config apis on a bucket

ebozduman:
This PR brings bucket encryption control thru Python SDK apis;

- put_bucket_encryption
- get_bucket_encryption
- delete_bucket_encryption

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -26,9 +26,7 @@ s3Client = Minio('s3.amazonaws.com',
                  secure=True)
 ```
 
-
-
-| Bucket operations                                     | Object operations                                       | Presigned operations                              | Bucket policy/notification operations                               |
+| Bucket operations                                     | Object operations                                       | Presigned operations                              | Bucket policy/notification/encryption operations                    |
 |:------------------------------------------------------|:--------------------------------------------------------|:--------------------------------------------------|:--------------------------------------------------------------------|
 | [`make_bucket`](#make_bucket)                         | [`get_object`](#get_object)                             | [`presigned_get_object`](#presigned_get_object)   | [`get_bucket_policy`](#get_bucket_policy)                           |
 | [`list_buckets`](#list_buckets)                       | [`put_object`](#put_object)                             | [`presigned_put_object`](#presigned_put_object)   | [`set_bucket_policy`](#set_bucket_policy)                           |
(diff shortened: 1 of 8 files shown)


docs/API.md
```diff
@@ -604,9 +613,93 @@ for event in events:
...
+}
+try:
+    minioClient.put_bucket_encryption('mybucket', ENC_CONFIG)
+except ResponseError as err:
+    print(err)
+```
+
+<a name="delete_bucket_encryption"></a>
+
+### remove default encryption configuration on a bucket
```
vadmeste: This needs to be consistent with others titles

examples/get_bucket_encryption.py
```diff
@@ -0,0 +1,32 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Note: YOUR-ACCESSKEYID, YOUR-SECRETACCESSKEY and my-bucketname are
+# dummy values, please replace them with original values.
+
+from minio import Minio
+from minio.error import ResponseError
+
+CLIENT = Minio('s3.amazonaws.com',
```
vadmeste: All capital letters usually means a constant, I think you should replace 'CLIENT' with 'client'

examples/remove_bucket_encryption.py
```diff
@@ -0,0 +1,32 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Note: YOUR-ACCESSKEYID, YOUR-SECRETACCESSKEY and my-bucketname are
+# dummy values, please replace them with original values.
+
+from minio import Minio
+from minio.error import ResponseError
+
+CLIENT = Minio('s3.amazonaws.com',
```
vadmeste: All capital letters usually means a constant, I think you should replace 'CLIENT' with 'client'

examples/set_bucket_encryption.py
```diff
@@ -0,0 +1,42 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Note: YOUR-ACCESSKEYID, YOUR-SECRETACCESSKEY and my-bucketname are
+# dummy values, please replace them with original values.
+
+from minio import Minio
+from minio.error import ResponseError
+
+CLIENT = Minio('s3.amazonaws.com',
```
vadmeste: All capital letters usually means a constant, I think you should replace 'CLIENT' with 'client'

minio/xml_marshal.py
```diff
@@ -29,9 +29,48 @@
 import io
 
 from xml.etree import ElementTree as s3_xml
+from collections import defaultdict
 
 _S3_NAMESPACE = 'http://s3.amazonaws.com/doc/2006-03-01/'
 
+def xml_to_dict(in_xml):
```
vadmeste: This function does not seem to be used

minio/xml_marshal.py
```diff
@@ -29,9 +29,48 @@
...
+    # Converts xml to dict
+    elem = s3_xml.XML(in_xml)
+    return etree_to_dict(elem)
+
+def etree_to_dict(elem):
```
vadmeste: This function does not seem to be used

minio/api.py
```diff
@@ -578,6 +578,73 @@ def remove_all_bucket_notification(self, bucket_name):
...
+        result = ''
+        is_valid_bucket_name(bucket_name, False)
+        try:
+            response = self._url_open(
+                "GET",
+                bucket_name=bucket_name,
+                query={"encryption": ""}
+            )
+            result = response.data.decode('utf-8')
+        except ResponseError as err:
```
vadmeste: I think we should not catch any exception here

docs/API.md
ebozduman: ok

examples/get_bucket_encryption.py
ebozduman: I have Python lint running and that was the recommendation. 
I thought we follow and adjust our code according to lint warnings. 
I'll change them back to lower case.
Could you also please confirm that we don't follow Python lint recommendations?
Thanks

examples/remove_bucket_encryption.py
ebozduman: ok

examples/set_bucket_encryption.py
ebozduman: ok

minio/xml_marshal.py
ebozduman: You're right. I was supposed to use it to return a dictionary instead of xml from `get_bucket_encryption` api.
I'll fix it.

minio/xml_marshal.py
ebozduman: This is the same issue with your previous comment.
`xml_to_dict` calls `etree_to_dict` and I am going to call `xml_to_dict` from `get_bucket_encryption`.

minio/api.py
ebozduman: Sure, I'll remove it.

ebozduman:
@vadmeste,
What version of Python did you use?

kannappanr:
ping @vadmeste @sinhaashish

PR #856, opened 2020-02-19
Update MANIFEST.in to generate a complete sdist

dlax:
We first include all files needed to run the tests from a sdist.
Previously, I get 4 errors and 4 failures due to files not found by
tests.

We also prune files which do not seem needed when using the sdist
package.

The correctness of the update MANIFEST.in can be verify using
check-manifest.


dlax:
> Please fix conflicts

Done.

Issue #857, opened 2020-02-28
AWS IAM Credential using subscripts on not subscriptable objects


febg:
When trying to use the IamEc2MetaData provider like this:
```
from minio import Minio
from minio.credentials import IamEc2MetaData, Credentials

# Initialize Minio with IamEc2MetaData default credentials object
client = Minio('s3.amazonaws.com',
               credentials=Credentials(provider=IamEc2MetaData()))


buckets = client.list_buckets()
```
I got the following error:
```
Traceback (most recent call last):
  File "test.py", line 10, in <module>
    client.list_buckets()
  File "/usr/local/lib/python2.7/site-packages/minio/api.py", line 411, in list_buckets
    raise AccessDenied(response)
minio.error.AccessDenied: AccessDenied: message: Access Denied
```

After doing some digging on why my credentials were not working I found the root cause:

IamEc2MetaData provider is trying to access http response data and status using subscripts throwing the following error:
```
TypeError: 'HTTPResponse' object is not subscriptable
```

balamurugana:
Fixed by https://github.com/minio/minio-py/pull/858

PR #858, opened 2020-02-28
Fixes not subscriptable objects

febg:
Fixes issue #857 

As the issue explains, IamEc2MetaData attempts to access http response data and status using subscripts like the example shown below:

```
if res['status'] != 200:
                return []
```

This throws an exception because the `'HTTPResponse' object is not subscriptable`.

### DIFF
--- a/minio/credentials/aws_iam.py
+++ b/minio/credentials/aws_iam.py
@@ -47,7 +47,7 @@ class IamEc2MetaData(Provider):
         except:
             return None
         creds = res.data
-        return creds.decode("utf-8")
+        return creds.decode("utf-8").split('\n')
 
     def request_cred(self, creds_name):
         url = self._endpoint + self.iam_security_creds_path + "/" + creds_name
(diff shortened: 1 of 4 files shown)


minio/credentials/aws_iam.py
```diff
@@ -42,32 +42,31 @@ def request_cred_list(self):
...
-            if res['status'] != 200:
-                return []
+            if res.status != 200:
+                return None
         except:
-            return []
-        creds = res['data'].split('\n')
-        return creds
+            return None
+        creds = res.data
```
vadmeste: Could we trim any '\n' or white space that we can find ? I guess that old logic of the code is there for a reason.

febg: The `split(\n)` was used in case more than one role name is returned  by AWS metadata service. However, according to AWS docs an EC2 machine can only have one role attached.

I added the `split(\n)` back just in case!

febg:
@harshavardhana waiting on your review.

PR #862, opened 2020-03-12
Add an option to connect to self signed servers

kabab:
Add an option `cert_insecure` to the Minio constructor, in order to connect to hosts with self signed and insecure certificate

Example:
```
minioClient  = Minio(endpoint='minio-stg.example.ma', 
                                    access_key='access_key',
                                    secret_key='secret_key', 
                                    secure=True,
                                    **cert_insecure**=True)
```

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -144,6 +144,8 @@ class Minio(object):
          location discovery.
     :param timeout: Set this value to control how long requests
          are allowed to run before being aborted.
+    :param cert_insecure: Set this value if you are using self signed certificate.
+        Default is False
     :return: :class:`Minio <Minio>` object
 
     **NOTE on concurrent usage:** The `Minio` object is thread safe when using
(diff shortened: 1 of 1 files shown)


vadmeste:
Thanks for the PR @kabab. 

Actually if SSL_CERT_FILE is set to the path of the public self signed certificate, then minio-py will just trust it. `cert_insecure` is probably not needed here

kabab:
@vadmeste thank you for the review, I agree with you, but sometimes it more flexible to have an option, it's more like using `curl` with `--insecure` option, and also you can use cert files with `curl` , it's an optional option that doesn't break the api.

harshavardhana:
> @vadmeste thank you for the review, I agree with you, but sometimes it more flexible to have an option, it's more like using `curl` with `--insecure` option, and also you can use cert files with `curl` , it's an optional option that doesn't break the API.

@kabab that is an overkill this is an SDK, not a command-line tool like `curl` we cannot make our SDK . `SSL_CERT_FILE` is meant for this reason alone.

kannappanr:
Clsoing this PR based on the above comments

PR #867, opened 2020-04-04
cElementTree has been deprecated since Python 3.3 and removed in Python 3.9

### LINKED ISSUE #863: xml.etree.cElementTree was removed in Python 3.9
There are instances in the code base where cElementTree is used. This was deprecated and removed in Python 3.9.

Ref : https://github.com/python/cpython/pull/19108
Slightly related : https://github.com/minio/minio-py/issues/822

tirkarthi:
Fixes #863

### DIFF
--- a/minio/compat.py
+++ b/minio/compat.py
@@ -61,9 +61,6 @@ if _is_py2:
 
     ## Add missing imports
     basestring = basestring
-
-    from xml.etree import cElementTree
-    from xml.etree.cElementTree import ParseError
 elif _is_py3:
     from queue import Queue
     queue = Queue
@@ -93,9 +90,6 @@ elif _is_py3:
     bytes = bytes
     range = range
     str = str
-
-    from xml.etree import ElementTree as cElementTree
-    from xml.etree.ElementTree import ParseError
 
 numeric_types = (int, long, float)
 
--- a/minio/compat.py
+++ b/minio/compat.py
@@ -61,6 +61,9 @@ if _is_py2:
 
     ## Add missing imports
     basestring = basestring
+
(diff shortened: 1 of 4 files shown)


harshavardhana:
@tirkarthi this is an incorrect fix

tirkarthi:
I tried using compat since it had python 2/3 related imports. Can you be specific about the approach needed please?

harshavardhana:
If celementTree is gone it should be completely remove and replaced with another XML alternative. The current fix is not doing anything other than just moving imports. 
@tirkarthi

harshavardhana:
The correct fix is just replace all cElementTree with ElementTree @tirkarthi

tirkarthi:
The project seems to support Python 2 and 3. Python 2 had ElementTree and cElementTree (accelerated version). Python 3 has accelerated version by default as ElementTree. Hence cElementTree was just an alias that was deprecated and removed. Always using ElementTree is a solution and more simple but it might cause slowdown in Python 2. If it's okay I can change it to that.

harshavardhana:
> The project seems to support Python 2 and 3. Python 2 had ElementTree and cElementTree (accelerated version). Python 3 has accelerated version by default as ElementTree. Hence cElementTree was just an alias that was deprecated and removed. Always using ElementTree is a solution and more simple but it might cause slowdown in Python 2. If it's okay I can change it to that.

That is not a problem @tirkarthi

tirkarthi:
Thanks @harshavardhana for the clarification. I have reverted the previous approach and made the suggested changes.

tirkarthi:
I am not sure if the testcase failures are related to this PR.

PR #868, opened 2020-04-04
Add read_bytes to put_object function

includeamin:
In a case of uploading the file to API with an HTTP request. it's required to be directly uploaded from API to Minio without saving the file as a temporary file.
for the small file, like Photos, partial files and etc.
for example :
``` python
from minio import Minio
from fastapi import FastAPI, File, UploadFile
from uvicorn import run

client = Minio("0.0.0.0:32768", secure=False, secret_key="minioadmin", access_key="minioadmin")

app = FastAPI()


@app.post("/")
async def index(file: UploadFile = File(...)):
    data = file.file.read()
    size = len(data)
    result = client.put_object("media-service", file.filename, data, size, read_bytes=True)

    return result


if __name__ == '__main__':
    run(app, host='0.0.0.0', port=3004)

```
``` sh
curl -X POST "http://0.0.0.0:3004/" -H "accept: application/json" -H "Content-Type: multipart/form-data" -F "file=@ffmpegX.dmg"
```

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -928,7 +928,7 @@ class Minio(object):
     def put_object(self, bucket_name, object_name, data, length,
                    content_type='application/octet-stream',
                    metadata=None, sse=None, progress=None,
-                   part_size=DEFAULT_PART_SIZE):
+                   part_size=DEFAULT_PART_SIZE,read_bytes=False):
         """
         Add a new object to the cloud storage server.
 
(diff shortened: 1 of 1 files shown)


harshavardhana:
This really doesn't make sense the code you pasted is wrong @includeamin you could have passed the entire file pattern to put_object rather than doing a file.read() yourself.

PR #869, opened 2020-04-05
fix github actions error while checking out

kannappanr:
using the fix proposed in https://github.com/actions/checkout/issues/23


harshavardhana:
build issues indicate that we have introduced a bug on windows when deleting objects @kannappanr

ebozduman:
This build issue seems to have started after Feb 19th (build checks worked for PR#856 the last time) and before Apr 3rd with my PR#866.

PR #870, opened 2020-04-06
Adding content_hash_hex parameter to signers.presign_v4

vsoch:
I have the following use case:

 - I'm deploying Minio alongside a server (uwsgi with Django, that's not important)
 - A command line client, [scs-library-client](https://github.com/sylabs/scs-library-client/blob/30f9b6086f9764e0132935bcdb363cc872ac639d/client/push.go) through Singularity is handling parsing an object to ask for mutlipart upload presigned URLs from my server. It conforms to the S3 API for Multipart upload.
 - This means that my server has endpoints to Start a multipart upload, and then a makes subsequent requests to retrieve a pre-signed URL, one for each part.
 - The Singularity client then uses the signed URL to interact with Minio (for those wondering this does mean that the external URL (127.0.0.1:9000) is different from the internal one (minio:9000) and to get around this I needed to make two clients, one for internal interaction (starting the multipart upload) and the other would have been used to generated the signed urls but I wound up just using it to create a prefix, I can elaborate more on this if anyone is interested).
 - After upload, Singularity again pings the server with a list of parts (numbers and ETags) that are then used for the (internal) s3 client to complete_multipart_upload.

The reason for this PR comes down to when Singularity uses the signed URL. I was using the signed url generated by the s3 client, that resulted in a signature mismatch. I next used the same flow that I saw for my working (single) PUT object request (also a signed URL from the Singularity client) that generates a URL and then uses signers.presign_v4. This continually failed. I tried about 1000 things, but the trick was that I needed to give the sha256 sum provided by the Singularity client to be included in the signature generation. Since I wasn't able to specify it for presign_v4, I essentially reproduced the function and used it as a variable. The error disappeared and my chunks sped through my localhost like a chicken with it's butt on fire! :chicken: :fire: 

So - this small fix will allow for an interested user that wants to use the minio-py client to generate a presigned url in the case that a server is making a request on behalf of a command line client (where the object to be uploaded lives and is not accessible to the server generating the URLs). The command line client sends the sha256sum to the server (which is not able to calculate it).  While I think it would be great for minio-py to have a fully supported function to generate these presigned URLS more easily (akin to how you can generate one for a single PUT request), in the meantime this small change to allow specifying the content_hash_hex will be useful to users that have similar use case to me. For future reference, the code I am referring to is https://github.com/singularityhub/sregistry/pull/298.

Signed-off-by: vsoch <vsochat@stanford.edu>


vsoch:
hey maintainers! It looks like this is some kind of race condition? Or at least the failed RUN is failing because it's reporting a bucket is not empty. Let me know if there is something I can do on my end - my changes to the function should not have led to any functional changes in the code as the default for my parameter is None, and it gets set to exactly what it was before.

harshavardhana:
> hey maintainers! It looks like this is some kind of race condition? Or at least the failed RUN is failing because it's reporting a bucket is not empty. Let me know if there is something I can do on my end - my changes to the function should not have led to any functional changes in the code as the default for my parameter is None, and it gets set to exactly what it was before.

Don't worry about it we need to fix the Windows CI build

minio/signer.py
```diff
@@ -74,6 +74,7 @@ def presign_v4(method, url, credentials,
     :param expires: final expiration of the generated URL. Maximum is 7days.
     :param response_headers: Specify additional query string parameters.
     :param request_date: the date of the request.
+    :param content_hash_hex: the sha256sum of the payload if already calcualted.
```
harshavardhana, line 77: can we call this `content_sha256_hex` instead to be more accurate?

vsoch: Sure - is that the only kind that it can be though? Note that if `X-Amz-Algorithm` is something different, or another hash header is provided to sign, why would the variable restrict to sha256? note that the variable already existed in the function and had the name `content_hash_hex` which is why I kept it like that.

harshavardhana: Yes in S3 that is always going to be sha256 unless entire AWS S3 changes.

harshavardhana: when that time comes we need to write a new SDK @vsoch

vsoch: haha okay, let me change that quickly then :)

harshavardhana: So to keep it generic we already take headers so if that covers it, then we don't need additional param - just honor `x-amz-content-sha256 ` from input headers.

vsoch:
> So to keep it generic we already take headers so if that covers it, then we don't need additional param - just honor x-amz-content-sha256 from input headers.

@harshavardhana I noticed that and tried supplying the same content hash via the headers variable, this resulted in the invalid signature still. I haven't looked closely at `generate_canonical_request` but it takes in `content_sha256` separately so it must be handled differently, or at least having it provided alongside a header with the correct sha256 somehow adds an incorrect value.

harshavardhana:
> @harshavardhana I noticed that and tried supplying the same content hash via the headers variable, this resulted in the invalid signature still. I haven't looked closely at `generate_canonical_request` but it takes in `content_sha256` separately so it must be handled differently, or at least having it provided alongside a header with the correct sha256 somehow adds an incorrect value.

The reason is that the code should look for `x-amz-content-sha256` present on headers and use that instead

minio/signer.py
```diff
@@ -144,7 +147,7 @@ def presign_v4(method, url, credentials,
                                                    new_parsed_url,
                                                    headers_to_sign,
                                                    signed_headers,
-                                                   content_hash_hex)
+                                                   content_sha256_hex)
```
harshavardhana, line 150: This is all there is to add `headers['x-amz-content-sha256']` and also make sure to purge it from the headers after

```
delete(headers, 'x-amz-content-sha256')
```

vsoch: Gotcha! I'll make these changes, likely tomorrow.

vsoch: @harshavardhana one concern that I have is with casing - how do we handle both cases when the user might do:

```python
headers={"X-Amz-Content-Sha256": sha256},
```
versus what you specified:

```python
headers={"x-amz-content-sha256": sha256},
```
We would need to parse through them and make all lower-case every time, which I'm not sure would make sense - the ones that are going into `query` to be signed have variable casing:

```
query["X-Amz-Algorithm"] = _SIGN_V4_ALGORITHM
```

On the other hand, to just expose the variable, we don't need this extra parsing and it just gets added as part of the signature to sign:

```python
    canonical_request = generate_canonical_request(
        method, new_parsed_url, headers_to_sign, signed_headers, content_hash_hex
    )
```
I see where you are coming from in not wanting to add extra variables, but if that is the best practice, then why is the content_hash_hex that currently exists getting special treatment, in terms of being it's own variable to pass into the second function? I think the approach to add it cleanly as a variable (with default as None) would both avoid the capitalization issues or need for extra parsing, and also already match the coding style (having it as a parameter) that already exists. Let me know your thoughts.

harshavardhana: That is easy every header set here is ultimately canonicalized so we look for lower-cased form and purge 


```py
>>> content_sha256=None
>>> for k in kv.keys():
...     if k.lower() == 'x-amz-content-sha256':
...             content_sha256 = kv.get(k)
...             kv.pop(k)
... 
'help'
>>> print(content_sha256)
help
>>> 
```

vsoch: Where exactly does this happen? So where would you like the check to be? I'm not following because my current usage has the headers passed in as a dictionary (with lower and uppercase) and this quickly gets passed into `get_signed_headers` that does handle the lowercase, but also turns it into a list:

```
def get_signed_headers(headers):
    """
    Get signed headers.

    :param headers: input dictionary to be sorted.
    """
    signed_headers = []
    for header in headers:
        signed_headers.append(header.lower().strip())
    return sorted(signed_headers)
```
so we would need to add an extra loop (as you've shown) to iterate through the dictionary of headers before that (in parse_v4) for this to work. Is this the approach that you would like?

harshavardhana: You need to add this in the presigned_v4 code @vsoch - if its confusing let me send the PR.

minio/signer.py
```diff
@@ -92,8 +92,13 @@ def presign_v4(method, url, credentials,
     if request_date is None:
         request_date = datetime.utcnow()
 
-    parsed_url = urlsplit(url)
     content_hash_hex = _UNSIGNED_PAYLOAD
+    for k in headers.keys():
+        if k.lower() == 'x-amz-content-sha256':
+            content_hash_hex = headers.get(k, _UNSIGNED_PAYLOAD)
+            del headers[k]
```
vsoch, line 99: The one tweak we could do here is to break when we find it, but it's not absolutely necessary. The size of headers will never be very large!

minio/signer.py
```diff
@@ -144,7 +149,7 @@ def presign_v4(method, url, credentials,
                                                    new_parsed_url,
                                                    headers_to_sign,
                                                    signed_headers,
-                                                   content_hash_hex)
+                                                   content_sha256_hex)
```
vsoch, line 152: one bug here!

minio/signer.py
vsoch: I added the break.

vsoch:
The one last bit I'd want to ask about is where we can document this? If it's included in this repository I can add it here, otherwise you can point me to where a user would want to find this detail.

harshavardhana:
> The one last bit I'd want to ask about is where we can document this? If it's included in this repository I can add it here, otherwise you can point me to where a user would want to find this detail.

Just add it as a comment, we don't really want to encourage this as most of our users are not aware of AWS S3 behavior and we would like to keep that abstracted.

vsoch:
okay, I'll add a comment to before the loop. One second!

vsoch:
okay I added a comment, and I also added a very detailed commit message so it's present in the git history too (and unexpected source of goodness sometimes!)

harshavardhana:
ping @vadmeste

minio/signer.py
```diff
@@ -92,8 +92,15 @@ def presign_v4(method, url, credentials,
     if request_date is None:
         request_date = datetime.utcnow()
 
-    parsed_url = urlsplit(url)
+    # If a sha256sum is known, add to headers to include with signature
     content_hash_hex = _UNSIGNED_PAYLOAD
+    for k in headers.keys():
+        if k.lower() == 'x-amz-content-sha256':
+            content_hash_hex = headers.get(k, _UNSIGNED_PAYLOAD)
```
hardbyte, line 99: I'm rebasing some changes on-top of this and wondered about this snippet.

Is there is a reason to call `keys()` (given iterating on a dict yields the keys anyway), and secondly the `headers.get` method call has a default of `_UNSIGNED_PAYLOAD` which appears to be unreachable by definition?

vsoch: keys() (my two cents) is mostly helpful for the developer reader the code to see that we are iterating through a dictionary. And I agree given that we find k as a key, the second value is unreachable. I think @harshavardhana wrote this snippet so maybe he can comment. It might just be another message for the developer, in code, to hint at what the value should be?

harshavardhana: why would it be unreachable? @hardbyte the reason it is done to ensure that we are case insensitive, otherwise we may look at the key which may be defined with a different canonical representation in the input headers dictionary. At the same time, we don't change what user-provided as input. 

We can definitely remove the `header.keys()`

vsoch: I think what we might want instead is:
```python
content_hash_hex = headers.get(k) or _UNSIGNED_PAYLOAD
```
This will mean that if the key is set but the value is None, we default to the unsigned payload. Currently, if the header is set (and it's None or empty string) we then use that empty value. What would be the desired outcome?

vsoch: Another edge error would be having that header set, but with different capitalization. Here is a dummy example:

```
 for k in mydict.keys(): 
    if k.lower() == 'a': 
        print(mydict[k]) 

a
A
```
In this case we would get whichever is last in the matching set.

hardbyte: Thanks for revisiting this.

@harshavardhana the `_UNSIGNED_PAYLOAD` is unreachable as far as I can tell - as `k` is definitely in `headers` - the default value won't ever apply.

In the first snippet @vsoch posted the default value does get used if the user has passed in something falsey (e.g. empty string or None). Perhaps passing in a falsey header should just raise a `ValueError` instead though.

harshavardhana: > Thanks for revisiting this.
> 
> @harshavardhana the `_UNSIGNED_PAYLOAD` is unreachable as far as I can tell - as `k` is definitely in `headers` - the default value won't ever apply.
> 
> In the first snippet @vsoch posted the default value does get used if the user has passed in something falsey (e.g. empty string or None). Perhaps passing in a falsey header should just raise a `ValueError` instead though.

Feel free to change it to be more pythonic, with consideration of case sensitivity.

PR #872, opened 2020-04-07
honor sha256 set in headers for presign_v4

harshavardhana:


harshavardhana:
// cc @vsoch

minio/signer.py
```diff
@@ -86,14 +87,19 @@ def presign_v4(method, url, credentials,
     if headers is None:
         headers = {}
 
+    content_hash_hex = _UNSIGNED_PAYLOAD
+    for k in headers.keys():
+        if k.lower() == 'x-amz-content-sha256':
+            content_hash_hex = headers.get(k, _UNSIGNED_PAYLOAD)
+            del headers[k]
+
     if expires is None:
```
vsoch, line 96: This is exactly what I was asking - if you wanted this loop added in the function so we would need to parse over all of the headers. Should I update my PR for this so that I am given credit for the work?

harshavardhana: Yes, definitely @vsoch - these are minimal things parsing all headers are expected. Most of this is a niche requirement.

PR #875, opened 2020-04-08
Rename urlencode and urldecode to quote and unquote

hardbyte:
Quote and unquote are the names of the urllib functions being wrapped, this PR is a minor refactor to use those names.

Motivation is to avoid confusion when needing to use the real `urlencode` function (cf [comment](https://github.com/minio/minio-py/pull/874#discussion_r405801164)).

### DIFF
--- a/minio/compat.py
+++ b/minio/compat.py
@@ -41,10 +41,8 @@ if _is_py2:
     queue_empty = Empty
 
     from urllib import quote
-    _urlencode = quote
 
     from urllib import unquote
-    urldecode = unquote
 
     import urlparse
     urlsplit = urlparse.urlsplit
@@ -69,10 +67,8 @@ elif _is_py3:
     queue_empty = Empty
 
     from urllib.request import quote
-    _urlencode = quote
 
     from urllib.request import unquote
-    urldecode = unquote
 
     import urllib.parse
     urlsplit = urllib.parse.urlsplit
@@ -93,16 +89,19 @@ elif _is_py3:
 
 numeric_types = (int, long, float)
 
-def urlencode(resource):
+
+# Note earlier versions of minio.compat exposed urllib.quote as urlencode
+def _quote(resource):
     """
-    This implementation of urlencode supports all unicode characters
+    This implementation of urllib.quote supports all unicode characters
 
     :param: resource: Resource value to be url encoded.
     """
     if isinstance(resource, str):
-        return _urlencode(resource.encode('utf-8'))
+        return quote(resource.encode('utf-8'))
 
-    return _urlencode(resource)
+    return quote(resource)
+
 
 def queryencode(query):
     """
@@ -110,4 +109,4 @@ def queryencode(query):
 
     :param: query: Query value to be url encoded.
     """
-    return urlencode(query).replace('/', '%2F')
+    return _quote(query).replace('/', '%2F')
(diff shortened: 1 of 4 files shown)


minio/compat.py
```diff
@@ -93,21 +93,23 @@
 
 numeric_types = (int, long, float)
 
-def urlencode(resource):
+
```
Praveenrajmani, line 96: Can you leave a note for the `deprecation` as a comment.. Just to record the change?

minio/compat.py
```diff
@@ -41,10 +41,10 @@
     queue_empty = Empty
 
     from urllib import quote
-    _urlencode = quote
+    _quote = quote
 
     from urllib import unquote
-    urldecode = unquote
+    unquote = unquote
```
Praveenrajmani, line 47: This change does not make sense here

minio/compat.py
hardbyte: Sure - I'll add as a comment rather than in the docstring so it isn't visible to end users.

minio/compat.py
hardbyte: Eek - this diff makes that painfully obvious! I'll remove this line.

hardbyte:
@Praveenrajmani that makes sense - I just wanted to make minimal changes.

I'll quickly refactor with what I understood your idea to be.

minio/compat.py
Praveenrajmani: I still see this @hardbyte. Can we name it as `_unquote`

hardbyte: Done

hardbyte:
So all the tests pass if we delete the custom wrapper:

```python
def _quote(resource):
    """
    This implementation of urllib.quote supports all unicode characters

    :param: resource: Resource value to be url encoded.
    """
    if isinstance(resource, str):
        return quote(resource.encode('utf-8'))

    return quote(resource)
```

If we do this the `minio.compat` module can export `quote` and `unquote` without needing to alias either.

Praveenrajmani:
> So the remaining issue is whether `minio.compat` module should export the Python standard library `quote` as well as the minio wrapper for `quote` (which I've exposed as `_quote`):
> 
> ```python
> def _quote(resource):
>     """
>     This implementation of urllib.quote supports all unicode characters
> 
>     :param: resource: Resource value to be url encoded.
>     """
>     if isinstance(resource, str):
>         return quote(resource.encode('utf-8'))
> 
>     return quote(resource)
> ```
> 
> There is no overriding function for `unquote` so I assume that exposing the standard library version as `minio.compat.unquote` is fine.

@hardbyte If there are no any overriding definitions of `unquote`, then lets us directly use it from `urllib.request` instead of chaining it over through `.compat` lib

To make sure we should not leave the other contributors to go check for the overriding definition of `unquote` and figure out that it is actually pointing to the standard one :P

Otherwise the PR looks good to me.

hardbyte:
I don't think it is quite that simple because minio-py is still trying to support Python2 and I assumed having one sure place to import from (e.g. `minio.compat.unquote`) is the main idea.

`unquote` in Python 2 is found in [`urllib.unquote`](https://docs.python.org/2/library/urllib.html#urllib.unquote), in Python 3 it is found in [`urllib.parse.unquote`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.unquote) (in Python 3 `unquote` is also exported by `urllib.request`). I'll push a tweak to use the real source for Python3.

Praveenrajmani:
PTAL @sinhaashish

PR #893, opened 2020-04-13
Allow signing requests for services other than s3 in sign v4

hardbyte:
Refactors `sign_v4` to take an optional `service` argument making it possible to sign for services other than `"s3"`, e.g. to allow the client to craft an `AssumeRole` request using the `"STS"` service.

*Note this PR extracts just the signature changes from PR #874*

### DIFF
--- a/minio/signer.py
+++ b/minio/signer.py
@@ -215,7 +215,7 @@ def sign_v4(method, url, region, headers=None,
                                                parsed_url,
                                                headers_to_sign,
                                                signed_headers,
-                                               content_sha256,
+                                               content_sha256
                                                )
 
     string_to_sign = generate_string_to_sign(request_datetime, region,
--- a/minio/signer.py
+++ b/minio/signer.py
@@ -42,6 +42,12 @@ _SIGN_V4_ALGORITHM = 'AWS4-HMAC-SHA256'
 # Hardcoded S3 header value for X-Amz-Content-Sha256
 _UNSIGNED_PAYLOAD = u'UNSIGNED-PAYLOAD'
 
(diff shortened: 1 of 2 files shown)


hardbyte:
Could you please clarify? I've tested this against minio but I haven't tested against AWS STS - I can probably do that in the next 24 hours though.

The other [PR](https://github.com/minio/minio-py/pull/874) I've opened on top of this adds the STS assume role api to minio-py. The relevant use of the signature code is [here](https://github.com/minio/minio-py/blob/fae417d7f6d0ab52f397ec3e1d6622352b0b9374/minio/credentials/assume_role.py#L66-L76).

harshavardhana:
> @harshavardhana  - could you please clarify your feedback? I've tested this against minio but I haven't tested against AWS STS - I can probably do that in the next 24 hours though.
> 
> The other [PR](https://github.com/minio/minio-py/pull/874) I've opened on top of this adds the STS assume role api to minio-py. The relevant use of the signature code is [here](https://github.com/minio/minio-py/blob/fae417d7f6d0ab52f397ec3e1d6622352b0b9374/minio/credentials/assume_role.py#L66-L76).

Yes ideally the change should come in together with the AssumeRole PR because service type STS requires payload to have content-sha256 but you shouldn't set the x-amz-content-sha256 header. This is the AWS STS behavior. 

On MinIO we don't have a restriction on x-amz-content-sha256 to be absent for STS call. Because if the server calculates an incorrect signature this won't work expecting strict set of headers.

Ideally send this PR together with AssumeRole STS feature, so that the PR can be tested together.

PR #900, opened 2020-04-26
Avoid passing the client to the assume role provider

Praveenrajmani:
Instead, have a callback with arguments (Merely a design fix)

### DIFF
--- a/examples/assume_role.py
+++ b/examples/assume_role.py
@@ -30,28 +30,26 @@
 #   those allowed by the policy of the role that is being assumed.
 # - YOUR-ACCESSKEYID and YOUR-SECRETACCESSKEY are
 #   dummy values, please replace them with original values.
-# - To use minio with AWS, the `Minio` client that is passed to the
-#   AssumeRoleProvider must have the endpoint 'sts.amazonaws.com', and the
-#   RoleARN argument must be provided.
-
+# - To use MinIO with AWS, pass the function - `get_assume_role_creds`
+#   as a lambda to the AssumeRoleProvider
 from minio import Minio
 from minio.credentials import AssumeRoleProvider, Credentials
 
-client = Minio('localhost:9000',
+client = Minio('s3.amazonaws.com',
                access_key='YOUR-ACCESSKEYID',
-               secret_key='YOUR-SECRETACCESSKEY'
+               secret_key='YOUR-SECRETKEY'
                )
 
-restricted_upload_policy = """{
+_RESTRICTED_UPLOAD_POLICY = """{
   "Version": "2012-10-17",
   "Statement": [
     {
       "Action": [
-        "s3:PutObject"
+        "s3:*"
       ],
       "Effect": "Allow",
       "Resource": [
-        "arn:aws:s3:::uploads/2020/*"
+        "arn:aws:s3:::my-bucket/*"
       ],
       "Sid": "Upload-access-to-specific-bucket-only"
     }
@@ -60,7 +58,7 @@ restricted_upload_policy = """{
 """
 
 credentials_provider = AssumeRoleProvider(
-    client, Policy=restricted_upload_policy)
+    get_assume_role_creds=lambda: client.get_assume_role_creds(policy=_RESTRICTED_UPLOAD_POLICY))
 temp_creds = Credentials(provider=credentials_provider)
 
 # User can access the credentials for e.g. serialization
@@ -69,4 +67,11 @@ print(temp_creds.get().access_key)
 print(temp_creds.get().secret_key)
 
 # Initialize Minio client with the temporary credentials
-restricted_client = Minio('localhost:9000', credentials=temp_creds)
+restricted_client = Minio('s3.amazonaws.com', credentials=temp_creds)
+
+# Get a full object.
+
+data = restricted_client.get_object('my-bucket', 'my-object')
+with open('/tmp/testfile', 'wb') as file_data:
+    for d in data.stream(32*1024):
+        file_data.write(d)
--- a/minio/api.py
+++ b/minio/api.py
@@ -46,7 +46,7 @@ import urllib3
 # Internal imports
 from . import __title__, __version__
 from .compat import range  # pylint: disable=redefined-builtin
-from .compat import basestring, quote, urlsplit
+from .compat import basestring, quote, urlencode, urlsplit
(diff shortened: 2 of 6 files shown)


examples/assume_role.py
```diff
@@ -60,7 +58,7 @@
 """
 
 credentials_provider = AssumeRoleProvider(
-    client, Policy=restricted_upload_policy)
+    get_assume_role_creds=lambda: client._get_assume_role_creds(policy=restricted_upload_policy))
```
balamurugana, line 61: It is bad idea to use private method. You can make `get_assume_role_creds` public.

minio/api.py
```diff
@@ -1949,6 +1949,54 @@ def _get_bucket_location(self, bucket_name):
...
+        role_session_name = "anything" if role_session_name is None else role_session_name
+
+        method = 'POST'
+        region = 'us-east-1'
+
+        query = {
+            "Action": "AssumeRole",
+            "Version": "2011-06-15",
+            "RoleArn": role_arn,
+            "RoleSessionName": role_session_name,
```
balamurugana, line 1964: You could use it directly

```python
"RoleArn": role_arn or "arn:xxx:xxx:xxx:xxxx",
"RoleSessionName": role_session_name or "anything",
```

minio/api.py
```diff
@@ -1949,6 +1949,54 @@ def _get_bucket_location(self, bucket_name):
             return 'eu-west-1'
         return location
 
+    def _get_assume_role_creds(self, role_arn=None, role_session_name=None, policy=None, duration_seconds=None):
```
balamurugana, line 1952: 1. Make this method public.
2. `role_*` is too verbose. you could have `arn` and `session_name`
3. I think `duration` is good enough than `duration_seconds`
4. Avoid `'` usage instead use `"` which is more standard.

minio/api.py
```diff
@@ -1949,6 +1949,54 @@ def _get_bucket_location(self, bucket_name):
...
+            "RoleArn": role_arn,
+            "RoleSessionName": role_session_name,
+        }
+
+        # Add optional elements to the request
+        if policy is not None:
+            query["Policy"] = policy
+
+        if duration_seconds is not None:
+            query["DurationSeconds"] = str(duration_seconds)
```
balamurugana, line 1972:
```python
if policy:
    query["Policy"] = policy

if duration_seconds:
    query["DurationSeconds"] = {0}.format(duration_seconds)
```

minio/api.py
```diff
@@ -1949,6 +1949,54 @@ def _get_bucket_location(self, bucket_name):
             return 'eu-west-1'
         return location
 
+    def _get_assume_role_creds(self, role_arn=None, role_session_name=None, policy=None, duration_seconds=None):
+
+        role_arn = "arn:xxx:xxx:xxx:xxxx" if role_arn is None else role_arn
+        role_session_name = "anything" if role_session_name is None else role_session_name
+
+        method = 'POST'
+        region = 'us-east-1'
```
balamurugana, line 1958: You can use these values directly than defining here.

minio/api.py
```diff
@@ -1949,6 +1949,54 @@ def _get_bucket_location(self, bucket_name):
...
+
+        # Add optional elements to the request
+        if policy is not None:
+            query["Policy"] = policy
+
+        if duration_seconds is not None:
+            query["DurationSeconds"] = str(duration_seconds)
+
+        url = self._endpoint_url + "/"
+        content = urlencode(query)
```
balamurugana, line 1975: Use these values directly.

minio/api.py
```diff
@@ -1949,6 +1949,54 @@ def _get_bucket_location(self, bucket_name):
...
+        signed_headers = sign_v4(method, url, region, headers,
+                                 self._credentials,
+                                 content_sha256=content_sha256_hex,
+                                 request_datetime=datetime.utcnow(),
+                                 service_name='sts'
+                                 )
+        response = self._http.urlopen(method, url,
+                                      body=content,
+                                      headers=signed_headers,
+                                      preload_content=True)
```
balamurugana, line 1992: Use below style is more standard.
```python
signed_headers = sign_v4(
    method, url,
    region,
    headers,
    self._credentials,
    content_sha256=content_sha256_hex,
    request_datetime=datetime.utcnow(),
    service_name='sts',
)
response = self._http.urlopen(
    method,
    url,
    body=content,
    headers=signed_headers,
    preload_content=True,
)
````

File: minio/parsers.py [hunk 8]
```diff
@@ -435,3 +436,26 @@ def parse_multi_object_delete_response(data):
...
+    """
+    Parser for assume role response.
+
+    :param data: XML response data for STS assume role as a string.
+    :return: A 2-tuple containing:
+        - a :class:`~minio.credentials.Value` instance with the temporary credentials.
+        - A :class:`DateTime` instance of when the credentials expire.
+    """
+    root = ElementTree.fromstring(data)
+    credentials_elem = root.find("sts:AssumeRoleResult", _XML_NS).find(
```
balamurugana, line 451: just use `credentials` or `element` here.

minio/credentials/assume_role.py
```diff
@@ -21,71 +21,22 @@
...
-
     # AWS STS support GET and POST requests for all actions. That is, the API does not require you to
     # use GET for some actions and POST for others. However, GET requests are subject to the limitation
     # size of a URL; although this limit is browser dependent, a typical limit is 2048 bytes. Therefore,
     # for Query API requests that require larger sizes, you must use a POST request.
-    method = 'POST'
 
-    def __init__(self, mc, RoleArn=None, RoleSessionName=None, Policy=None, DurationSeconds=None):
-        self._minio_client = mc
+    def __init__(self, get_assume_role_creds=None):
```
balamurugana, line 32: This should accept `callback_func` and its not an optional arg.

Praveenrajmani:
Addressed the comments @balamurugana. PTAL.

minio/api.py
```diff
@@ -45,8 +45,7 @@
 
 # Internal imports
 from . import __title__, __version__
-from .compat import range  # pylint: disable=redefined-builtin
```
balamurugana, line 48: If you remove this, `range` usage in this module does not use `xrange` in python2.

Praveenrajmani:
PTAL @balamurugana @harshavardhana

PR #901, opened 2020-04-26
refactor minio.credentials

balamurugana:
Depends on `Minio.get_assume_role_creds()` in https://github.com/minio/minio-py/pull/900


Makefile
```diff
@@ -1,6 +1,9 @@
 .PHONY: examples tests publish
 
 check:
+	@which pylint >/dev/null || pip install --user --upgrade pylint
+	@if python --version | grep -qi 'python 3'; then pylint --reports=no minio/credentials/*.py; fi
```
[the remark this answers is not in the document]
Praveenrajmani: Discussed this offline: Will make this global once all the other unit lint PRs are merged

minio/credentials/credentials.py
```diff
@@ -14,60 +14,61 @@
...
-        return self._forceRefresh or self._provider.is_expired()
+        self._value = None
+        self._validity = None
+
+    def get(self, force=False):
+        """Get credentials from provider if needed."""
+        if (
+                force or
+                not self._value or
+                (self._validity and self._validity < datetime.utcnow())
```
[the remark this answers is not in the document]
Praveenrajmani: Discussed this offline: This line - https://github.com/minio/minio-py/pull/901/files#diff-fad0ee1d2311a62dab3b9d3eb1894d1bR48 will set the timezone.

vadmeste:
@balamurugana could you explain what the API change done in this PR ? 

Also could you make sure that examples are updated ?

balamurugana:
> @balamurugana could you explain what the API change done in this PR ?
Below is the design change I proposed earlier and this PR is a refactor with this design.
```python
from abc import ABCMeta, abstractmethod
from datetime import datetimeclass Value():
    """
    Denotes credential values such as access key, secret key and session token.
    """
    def init(self, access_key, secret_key, session_token):
        self.access_key = access_key
        self.secret_key = secret_key
        self.session_token = session_tokenclass Provider():
    """Credential retriever."""
    metaclass = ABCMeta    @abstractmethod
    def retrieve(self) -> (Value, datetime):
        """Retrieve credential value and its expiry."""class Credentials():
    def __init__(self, provider):
        self._provider = provider
        self._value = None
        self._validity = None    def get(self, force=False):
        if (
                not self._value or
                (self._validity and self._validity < datetime.utcnow()) or
                force
        ):
            self._value, self._validity = self._provider.retrieva()
        return self._value
```


> Also could you make sure that examples are updated ?

Let me check once again.

minio/credentials/providers.py
```diff
@@ -0,0 +1,238 @@
...
+            try:
+                creds, expiry = provider.retrieve()
+                if creds:
+                    self._provider = provider
+                    return creds, expiry
+            except ValueError:
+                # ignore this error and try other providers.
+                pass
+
+        raise ValueError("no credentials retrieved")
```
vadmeste, line 76: Maybe here we should return the last encountered ValueError exception.. so at least the user can see at least a more self explaining error.

minio/api.py
```diff
@@ -178,10 +177,9 @@ def __init__(self, endpoint, access_key=None,
         self._credentials = credentials or Credentials(
             provider=Chain(
                 providers=[
-                    Static(access_key, secret_key, session_token),
+                    Static(Value(access_key, secret_key, session_token)),
```
vadmeste, line 180: Maybe it is enough if we can credentials in this form:

```
  Static(access_key, secret_key, session_token),
```

and create a Value instance inside Static init() function.

Let's review the API with @abperiasamy one last time.

minio/credentials/providers.py
balamurugana: Done

minio/api.py
balamurugana: I have changed to `Static(access_key, secret_key, session_token=None, expiry=None)`

minio/api.py
```diff
@@ -181,7 +180,6 @@ def __init__(self, endpoint, access_key=None,
                     Static(access_key, secret_key, session_token),
                     EnvAWS(),
                     EnvMinio(),
-                    IamEc2MetaData(),
```
vadmeste, line 184: Well, this is a breaking change, we should not forget about mentioning it when we release minio-py

balamurugana: Yes. I think we are covering in https://github.com/minio/minio-py/issues/902

kannappanr:
ping @Praveenrajmani

PR #906, opened 2020-05-04
fix pylint errors in compat.py

balamurugana:


johnnyvf24:
Hey! you guys broke my changes for https://github.com/minio/minio-py/pull/912 !? Without that single line change the minio client is incompatible with netapp storage grid. Can you please re-add something equivalent to 
```
 headers['X-Amz-Copy-Source'] = queryencode(
            object_source).replace('%2F', '/')
```?

I've been waiting forever for PyPi to have a release that works with netapp.

balamurugana:
@johnnyvf24 We have removed python 2 compatibility and `master` handles the case as per python 3. Please check https://github.com/minio/minio-py/blob/master/minio/helpers.py#L82

PR #907, opened 2020-05-04
Allow copy_object an optional url_encode parameter

johnnyvf24:
Hello,

I am currently using the minio python sdk for both a private minio instance and another s3 compatible service. Unfortunately the other service does not allow url-encoded values in the `x-amz-copy-source` http header. This pull request allows an optional flag to be included in `copy_object` that defines whether or not the value should be url encoded.

Thanks,
John

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -824,7 +824,8 @@ class Minio(object):
                                         sse=sse)
 
     def copy_object(self, bucket_name, object_name, object_source,
-                    conditions=None, source_sse=None, sse=None, metadata=None):
+                    conditions=None, source_sse=None, sse=None, metadata=None,
+                    url_encode=True):
         """
         Copy a source object on object storage server to a new object.
 
@@ -839,6 +840,8 @@ class Minio(object):
         supported CopyObject conditions.
         :param metadata: Any user-defined metadata to be copied along with
         destination object.
+        :param url_encode: A flag that defines whether or not the
+        object_source should be url encoded.
         """
         is_valid_bucket_name(bucket_name, False)
         is_non_empty_string(object_name)
@@ -864,7 +867,11 @@ class Minio(object):
             is_valid_sse_object(sse)
             headers.update(sse.marshal())
 
-        headers['X-Amz-Copy-Source'] = queryencode(object_source)
+        if url_encode:
+            headers['X-Amz-Copy-Source'] = queryencode(object_source)
+        else:
+            headers['X-Amz-Copy-Source'] = object_source
+
         response = self._url_open('PUT',
                                   bucket_name=bucket_name,
                                   object_name=object_name,


minio/api.py
```diff
@@ -824,7 +824,8 @@ def get_partial_object(self, bucket_name, object_name, offset=0, length=0,
                                         sse=sse)
 
     def copy_object(self, bucket_name, object_name, object_source,
-                    conditions=None, source_sse=None, sse=None, metadata=None):
+                    conditions=None, source_sse=None, sse=None, metadata=None,
+                    url_encode=True):
```
harshavardhana, line 828: This is not optional as per AWS S3 @johnnyvf24 - if this to make it work with an S3 compatible server the server should be fixed here.

harshavardhana: can you also state which is this S3 compatible service? do they GitHub issues?

johnnyvf24: The service is netapp. Would you happen to know of a work-around? Pardon me, I'm not familiar with the Spec, do you happen to have a link that defines it? Thanks

harshavardhana: There is no workaround @johnnyvf24 https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html

> x-amz-copy-source
> The name of the source bucket and key name of the source object, separated by a slash (/). Must be URL-encoded.

> Pattern: \/.+\/.+

harshavardhana: The server should be fixed here we cannot carry this baggage in our SDK

johnnyvf24: I see. I'll take it up with NetApp then

harshavardhana:
Will close this PR @johnnyvf24 as this is not actionable for us.

minio/api.py
johnnyvf24: Actually while looking at that same page via the examples it seems the `/` is not url-encoded. Not that you are wrong, just it seems odd to me.

harshavardhana: It is not path encoded it is query URL encoded style @johnnyvf24 - that is what they mean by URL encoded style.

harshavardhana: Those examples are just placeholders what we have implemented is the real server behavior.

johnnyvf24: Didn't notice that, but regardless, it seems this sdk is replacing all `/` with `%2F`? @harshavardhana

harshavardhana: Correct because python query encoding is broken @johnnyvf24

johnnyvf24: @harshavardhana after looking for examples online I think I have found the issue. It seems the individual components between the `/`s should be uri escaped but the slashes themselves should be untouched. As a basis for this claim, here is an old pull request that does exactly what we are talking about: https://github.com/aws/aws-sdk-js/pull/714/files .

The actual part that used to do the encoding is found here: https://github.com/aws/aws-sdk-js/blob/c2ad3ef8fd486bb58edc225f74152dda51f9f2fe/lib/util.js#L51

johnnyvf24: The line in `compat.py` referenced above replaces all the `/` in the string. I can confirm that when calling NetApp without the `%2F`s that the call works as intended

harshavardhana: Then please send a PR with the fix we shall take it in.

johnnyvf24: Will do! Not sure how the server side should be handled, so I'll just post one for this SDK.

PR #912, opened 2020-05-05
change encoding of x-amz-copy-source in copy_object

johnnyvf24:
Hello Minio Team,

Talked with @harshavardhana and it seems the encoding is a bit off for the minio python sdk. It seems the value of the `x-amz-copy-source` does not need to encode the individual `/` characters in a provided path.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -864,10 +864,9 @@ class Minio(object):
             is_valid_sse_object(sse)
             headers.update(sse.marshal())
         
-        # Uri-encode everything but the '/' character
-        object_source = queryencode(object_source).replace('%2F', '/')
         # URI encoded, except '/' as per AWS S3 requirement
-        headers['X-Amz-Copy-Source'] = queryencode(object_source).replace('%2F', '/')
+        headers['X-Amz-Copy-Source'] = queryencode(
+            object_source).replace('%2F', '/')
 
         response = self._url_open('PUT',
                                   bucket_name=bucket_name,
--- a/minio/api.py
+++ b/minio/api.py
@@ -866,7 +866,8 @@ class Minio(object):
         
         # Uri-encode everything but the '/' character
         object_source = queryencode(object_source).replace('%2F', '/')
-        headers['X-Amz-Copy-Source'] = object_source
+        # URI encoded, except '/' as per AWS S3 requirement
+        headers['X-Amz-Copy-Source'] = queryencode(object_source).replace('%2F', '/')
 
         response = self._url_open('PUT',
                                   bucket_name=bucket_name,
--- a/minio/api.py
+++ b/minio/api.py
@@ -863,8 +863,12 @@ class Minio(object):
         if sse:
             is_valid_sse_object(sse)
             headers.update(sse.marshal())
+        
+        # Uri-encode everything but the '/' character
+        copy_src = [ queryencode(part) for part in object_source.split('/') ]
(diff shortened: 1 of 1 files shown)


minio/api.py
```diff
@@ -863,8 +863,12 @@ def copy_object(self, bucket_name, object_name, object_source,
         if sse:
             is_valid_sse_object(sse)
             headers.update(sse.marshal())
+        
+        # Uri-encode everything but the '/' character
+        copy_src = [ queryencode(part) for part in object_source.split('/') ]
+        copy_src = '/'.join(copy_src)
+        headers['X-Amz-Copy-Source'] = copy_src
```
harshavardhana, line 870: @johnnyvf24 fix is very simple 

```
queryencode(object_source).replace('%2F', '/')
```

johnnyvf24: That is easier :+1:

johnnyvf24: do you know why the checks are failing?

harshavardhana: https://github.com/minio/minio-py/pull/912/checks?check_run_id=646883805 can be checked right here

minio/api.py
```diff
@@ -863,8 +863,11 @@ def copy_object(self, bucket_name, object_name, object_source,
         if sse:
             is_valid_sse_object(sse)
             headers.update(sse.marshal())
+        
+        # Uri-encode everything but the '/' character
+        object_source = queryencode(object_source).replace('%2F', '/')
+        headers['X-Amz-Copy-Source'] = object_source
```
harshavardhana, line 869:
```suggestion
        # URI encoded, except '/' as per AWS S3 requirement
        headers['X-Amz-Copy-Source'] = queryencode(object_source).replace('%2F', '/')
```

minio/api.py
johnnyvf24: I don't understand the error. It seems it is using a Makefile, but I don't see if it's a pylint failure, etc... Could you point me to a resource? Not sure why that one line would break the project.

harshavardhana: You should run `autopep8 --diff --exit-code *.py`

johnnyvf24:
Given my bad commits, I'd definitely compress them on merging.

vadmeste:
@johnnyvf24 I just tested copy_object() and it seems it works fine. Could you say what's specific in your case ?

johnnyvf24:
@vadmeste how did you test it? I have yet to test AWS, but based on their JS SDK, it seems that the actual slashes are not to be encoded. In my case I am using this python client for a minio instance and a netapp storagegrid S3 compatible service. This client has worked great for everything except for copy object. In my particular case I was trying to copy an object such as sample/path/to/object from one bucket to another in the netapp storage service. At first I thought it was a netapp problem but based on code I have seen from other clients I now think this is a minio-py difference in interpretation of implementation.

johnnyvf24:
@vadmeste any update on this?

harshavardhana:
> @johnnyvf24 I just tested copy_object() and it seems it works fine. Could you say what's specific in your case ?

The issue is with NetApp Storage Grid which is not forthcoming in its parsing of x-amz-copy-source here, although there is no need to convert / to be %2F it was implicitly done by minio-py which sort of breaks NetApp Storage Grid product. 

NOTE: without this change minio-py works fine with MinIO and AWS S3 because of both support %2F as well automatically. 

But it doesn't hurt to fix minio-py

Issue #934, opened 2020-06-29
AWS IAM provider fails to renew token upon expiration


febg:
The AWS IAM provider adds a 5 minute time delta to the actual expiration time of the token provided by AWS.

https://github.com/minio/minio-py/blob/99dd78b3a44a9ad2c5fb636b2c0de18ef7b3a82a/minio/credentials/providers.py#L236

This causes a 5 minute period of time in which the SDK is operating with expired credentials before asking for a new token.

harshavardhana:
> This causes 5 minute period of time in which the SDK is operating with expired credentials before asking for a new token.

The best way is to have something like a delta which negates not adds - look at the API here 

https://pkg.go.dev/github.com/minio/minio-go/v6@v6.0.57/pkg/credentials?tab=doc#Expiry.SetExpiration

```go
// SetExpiration sets the expiration IsExpired will check when called.
//
// If window is greater than 0 the expiration time will be reduced by the
// window value.
//
// Using a window is helpful to trigger credentials to expire sooner than
// the expiration time given to ensure no requests are made with expired
// tokens.
func (e *Expiry) SetExpiration(expiration time.Time, window time.Duration) {
	e.expiration = expiration
	if window > 0 {
		e.expiration = e.expiration.Add(-window)
	}
}
```

https://github.com/minio/minio-go/blob/51222d26253e90224f8b4f8ba9be9b7982e3ce53/pkg/credentials/iam_aws.go#L40

https://github.com/minio/minio-go/blob/51222d26253e90224f8b4f8ba9be9b7982e3ce53/pkg/credentials/iam_aws.go#L140

febg:
@harshavardhana I changed the time delta addition to a subtraction in PR #935.

PR #935, opened 2020-06-29
Removed iam time delta

febg:
Subtracts time delta from AWS IAM expiration time.

Tries to solve #934

### DIFF
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -185,7 +185,17 @@ class FileMinioClient(Provider):
 
 
 class IAMProvider(Provider):
-    """IAM EC2 credential provider."""
+    """
+        IAM EC2 credential provider.
+    
+        expiry_delta param is used to create a window to the token expiration time.
+        If expiry_delta is greater than 0 the expiration time will be reduced by the
+        delta value.
+
+        Using a delta value is helpful to trigger credentials to expire sooner than
+        the expiration time given to ensure no requests are made with expired token.
+
+    """
 
     def __init__(self,
                  endpoint=None,
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -187,13 +187,14 @@ class FileMinioClient(Provider):
 class IAMProvider(Provider):
     """
         IAM EC2 credential provider.
-    
-        expiry_delta param is used to create a window to the token expiration time.
-        If expiry_delta is greater than 0 the expiration time will be reduced by the
-        delta value.
 
-        Using a delta value is helpful to trigger credentials to expire sooner than
-        the expiration time given to ensure no requests are made with expired token.
+        expiry_delta param is used to create a window to the token
+        expiration time. If expiry_delta is greater than 0 the
+        expiration time will be reduced by the delta value.
+
+        Using a delta value is helpful to trigger credentials to
+        expire sooner than the expiration time given to ensure no
+        requests are made with expired token.
 
     """
 
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -187,7 +187,11 @@ class FileMinioClient(Provider):
 class IAMProvider(Provider):
     """IAM EC2 credential provider."""
 
-    def __init__(self, endpoint=None, http_client=None, expiry_time_delta=timedelta(seconds=0)):
+    def __init__(self,
+                 endpoint=None,
+                 http_client=None,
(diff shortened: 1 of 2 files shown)


minio/credentials/providers.py
```diff
@@ -187,7 +187,11 @@ def retrieve(self):
 class IAMProvider(Provider):
     """IAM EC2 credential provider."""
 
-    def __init__(self, endpoint=None, http_client=None):
+    def __init__(self,
+                 endpoint=None,
+                 http_client=None,
+                 expiry_delta=timedelta(seconds=0)):
```
garrettheel, line 193: given that a `timedelta` is mutable, it should not be used as a default value here. it would be better to set this to `None` and do `if expiry_delta is None: expiry_delta = ...`

harshavardhana: expiry_delta should be 10secs default and can be changed to a lower of higher value as necessary.

garrettheel: I'm just referring to https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments, not the value itself

febg: Gotcha, I changed the default argument value to be `None`. If that is the case we then set `expiry_time` to 10 secs

minio/credentials/providers.py
```diff
@@ -233,7 +241,7 @@ def retrieve(self):
             data["AccessKeyId"],
             data["SecretAccessKey"],
             session_token=data["Token"],
-        ), expiration + timedelta(minutes=5)
+        ), expiration - self._expiry_delta
```
vadmeste, line 244: Would you please add a comment to explain why we are negating here ?

minio/credentials/providers.py
```diff
@@ -187,7 +187,11 @@ def retrieve(self):
 class IAMProvider(Provider):
     """IAM EC2 credential provider."""
 
-    def __init__(self, endpoint=None, http_client=None):
+    def __init__(self,
+                 endpoint=None,
+                 http_client=None,
+                 expiry_delta=None):
```
vadmeste, line 193: Maybe it is not that useful to add expiry_delta parameter here. Let's negate 10 seconds by default unless we hear something from the community (I don't think we will)

harshavardhana: Yes it is 10sec inside the code @vadmeste as an if condition rather than defaulted at the constructor.

minio/credentials/providers.py
harshavardhana: We should bring the comment from minio-go

febg: This comment?
```
// SetExpiration sets the expiration IsExpired will check when called.
//
// If window is greater than 0 the expiration time will be reduced by the
// window value.
//
// Using a window is helpful to trigger credentials to expire sooner than
// the expiration time given to ensure no requests are made with expired
// tokens.
```

harshavardhana: Anything that would make sense to use this value

minio/credentials/providers.py
vadmeste: > Yes it is 10sec inside the code @vadmeste as an if condition rather than defaulted at the constructor.

I am just saying, let's not expose this variable to users. I don't see how this is useful. We can make it 10 seconds by default. (LGTM otherwise)

harshavardhana: It is necessary for situations where renewal can be aggressive or faster.  It also complies with minio-go behavior. So a configurable value is not a problem.

minio/credentials/providers.py
harshavardhana: @febg can you add the comment?

harshavardhana:
> AttributeError: module 'isort' has no attribute 'SortImports'


when did this start happening?

harshavardhana:
> module 'isort' has no attribute

https://github.com/timothycrosley/isort/issues/1273#issuecomment-654656067 @febg shall fix this in a separate PR hold on

vadmeste:
> > AttributeError: module 'isort' has no attribute 'SortImports'
> 
> when did this start happening?

Most likely we had opened PRs that successfully finished CI and we merged them after merging @balamurugana pylint PR

balamurugana:
A recent breakage of isort. We may need to upgrade packages as suggested.

harshavardhana:
> A recent breakage of isort. We may need to upgrade packages as suggested.

Yes sent another PR #937

PR #936, opened 2020-07-02
Allow presigned URL operations to specify an alternative endpoint_url

brianhelba:
This is useful in situations where the MinIO client accesses the server from a different endpoint URL than the end user who will consume the presigned URL. Such situations may occur in development, particularly with Docker Compose, or in production, particularly within a VPC or when connecting to a Kubernetes Service.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -1389,7 +1389,8 @@ class Minio:  # pylint: disable=too-many-public-methods
                       object_name,
                       expires=timedelta(days=7),
                       response_headers=None,
-                      request_date=None):
+                      request_date=None,
+                      endpoint_url=None):
         """
         Presigns a method on an object and provides a url
 
@@ -1412,6 +1413,9 @@ class Minio:  # pylint: disable=too-many-public-methods
         :params request_date: Optional request_date argument to
                               specify a different request date. Default is
                               current date.
+        :param endpoint_url: Optional endpoint_url argument to specify a
+                             different server URL, instead of getting the
+                             endpoint from this client.
         :return: Presigned put object url.
         """
         is_valid_bucket_name(bucket_name, False)
@@ -1424,9 +1428,12 @@ class Minio:  # pylint: disable=too-many-public-methods
                 ' {0} secs'.format(_MAX_EXPIRY_TIME))
 
         region = self._get_bucket_region(bucket_name)
-        endpoint_url = self._endpoint_url
-        if self._enable_s3_accelerate:
-            endpoint_url = self._accelerate_endpoint_url
+
+        if endpoint_url is None:
+            if self._enable_s3_accelerate:
+                endpoint_url = self._accelerate_endpoint_url
+            else:
+                endpoint_url = self._endpoint_url
 
         url = get_target_url(endpoint_url,
                              bucket_name=bucket_name,
@@ -1443,7 +1450,8 @@ class Minio:  # pylint: disable=too-many-public-methods
     def presigned_get_object(self, bucket_name, object_name,
                              expires=timedelta(days=7),
                              response_headers=None,
-                             request_date=None):
+                             request_date=None,
+                             endpoint_url=None):
         """
         Presigns a get object request and provides a url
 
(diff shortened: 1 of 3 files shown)


minio/api.py
```diff
@@ -1389,7 +1389,8 @@ def presigned_url(self, method,
                       object_name,
                       expires=timedelta(days=7),
                       response_headers=None,
-                      request_date=None):
+                      request_date=None,
+                      endpoint_url=None):
```
balamurugana, line 1393: Is it assumed the same credentials work for this endpoint? It just makes confusion for users. Can't you fix the issue at gateway side of minio server?

harshavardhana: Agreed @balamurugana these are ad-hoc use cases we shouldn't entertain. This library is simple as it is and serves it purpose. Generate presigned URL for the right URL with the right constructor is the way to go.

We should avoid cockpit-style options

brianhelba:
@balamurugana @harshavardhana Thanks for the feedback. I agree that adding lots of ad-hoc options to support niche use cases would be bad for the usability and maintainability of this library.

However, one issue is that [the call](https://github.com/minio/minio-py/pull/936/files#diff-6b286c0c99801437fe6764292b7b7469R1430):
```
region = self._get_bucket_region(bucket_name)
```
when run from a new `Minio` will often attempt an HTTP connection to the new endpoint. This will fail if the new server endpoint cannot be resolved from the caller. For example, the caller may **only** be able to reach the server at `minio:9000`, but needs to generate URLs for downloads from `minio.mysite.com`.

---

One workaround is something like:
```
client = Minio('minio:9000')
client.bucket_exists('my-bucket') # Populates the client._region_map cache

new_client = Minio('minio.mysite.com') # This can't actually make HTTP connections
new_client._set_bucket_region( # Populate new_client._region_map manually
    'my-bucket',
    client._get_bucket_region('my-bucket')
)

new_client.presigned_get_object('my-bucket', 'some-key') # Will not attempt any HTTP requests
```

The biggest problem with this (in addition to the complexity) is that it relies on private `_get_bucket_region` / `_set_bucket_region` methods. Would it be reasonable to have a new public way to specify the bucket region without making HTTP requests?

---

Even if it's not possible to resolve any of this upstream in `minio-py`, I wanted to at least make you aware of my use case here. I've gone ahead and implemented this in the downstream `django-minio-storage` library: https://github.com/py-pa/django-minio-storage/pull/97 (for MinIO support in Django).

harshavardhana:
> @balamurugana @harshavardhana Thanks for the feedback. I agree that adding lots of ad-hoc options to support niche use cases would be bad for the usability and maintainability of this library.
> 
> However, one issue is that [the call](https://github.com/minio/minio-py/pull/936/files#diff-6b286c0c99801437fe6764292b7b7469R1430):
> 
> ```
> region = self._get_bucket_region(bucket_name)
> ```
> 
> when run from a new `Minio` will often attempt an HTTP connection to the new endpoint. This will fail if the new server endpoint cannot be resolved from the caller. For example, the caller may **only** be able to reach the server at `minio:9000`, but needs to generate URLs for downloads from `minio.mysite.com`.
> 
> One workaround is something like:
> 
> ```
> client = Minio('minio:9000')
> client.bucket_exists('my-bucket') # Populates the client._region_map cache
> 
> new_client = Minio('minio.mysite.com') # This can't actually make HTTP connections
> new_client._set_bucket_region( # Populate new_client._region_map manually
>     'my-bucket',
>     client._get_bucket_region('my-bucket')
> )
> 
> new_client.presigned_get_object('my-bucket', 'some-key') # Will not attempt any HTTP requests
> ```
> 
> The biggest problem with this (in addition to the complexity) is that it relies on private `_get_bucket_region` / `_set_bucket_region` methods. Would it be reasonable to have a new public way to specify the bucket region without making HTTP requests?
> 

Just specify the region, it will not make requests. region request is attempted when there are not default regions set. MinIO access all regions by default so this way you can turn off the get_bucket_region requests.

> Even if it's not possible to resolve any of this upstream in `minio-py`, I wanted to at least make you aware of my use case here. I've gone ahead and implemented this in the downstream `django-minio-storage` library: [py-pa/django-minio-storage#97](https://github.com/py-pa/django-minio-storage/pull/97) (for MinIO support in Django).

Unfortunately this PR is not correct for our project, you may fork minio-py and keep the code - current contracts in APIs cannot be changed. We have rejected many such changes in the past in minio-go, minio-js projects.

harshavardhana:
Closing this PR as there are sufficient alternatives available to avoid bucket region lookup.

PR #937, opened 2020-07-08
fix: pylint version compatibility with isort

harshavardhana:


Makefile
```diff
@@ -1,28 +1,28 @@
 .PHONY: examples tests publish
 
 check:
-	@which pylint >/dev/null || pip install --user --upgrade pylint
-	@if python --version | grep -qi 'python 3'; then pylint --reports=no --score=no --disable=R0401 minio/*py; fi
-	@if python --version | grep -qi 'python 3'; then pylint --reports=no --score=no minio/credentials minio/select tests/functional; fi
+	@pip install --user --upgrade pylint==2.5.3
+	@if python --version | grep -qi 'python 3'; then ${HOME}/.local/bin/pylint --reports=no --score=no --disable=R0401 minio/*py; fi
+	@if python --version | grep -qi 'python 3'; then ${HOME}/.local/bin/pylint --reports=no --score=no minio/credentials minio/select tests/functional; fi
```
balamurugana, line 6: 1. Running pip install everything slows down for users. I think you could keep the check.
2. Referring ` ${HOME}/.local/bin/` is not needed so that the executable in the path will be used.

harshavardhana: > 2\. Referring ` ${HOME}/.local/bin/` is not needed so that the executable in the path will be used.

This is not in the PATH on GitHub CI @balamurugana 

> * Running pip install everything slows down for users. I think you could keep the check.

It is to ensure that we always install the right version for the users @balamurugana once it's installed its pretty quick it doesn't re-install.

balamurugana: For users:
1. I would recommend the check (I added the check purposefully when adding it) to avoid unnecessary pip lookup.
2. It is possible to have these tools in different path i.e. my local path may be different than `${HOME}`. Also these tools could have pre-installed in the OS. It is not a good idea to hard-code the path.

Github workflow:
1. Github workflow scripts need a fix if it is not in the path; but I don't think it a problem because they were working fine after `pip install` in the scripts.

Makefile
```diff
@@ -1,28 +1,28 @@
 .PHONY: examples tests publish
 
 check:
-	@which pylint >/dev/null || pip install --user --upgrade pylint
-	@if python --version | grep -qi 'python 3'; then pylint --reports=no --score=no --disable=R0401 minio/*py; fi
-	@if python --version | grep -qi 'python 3'; then pylint --reports=no --score=no minio/credentials minio/select tests/functional; fi
+	@pip install --user --upgrade pylint==2.5.3
```
balamurugana, line 4: 1. I don't think it is a good idea to stick to a version instead we need to install the latest for github ci.
2. Its not required to install `isort` specifically because pip install pylint automatically resolves it dependencies and we are good with the dependent isort.

Makefile
harshavardhana: They were not @balamurugana that's why I fixed them

harshavardhana: > * I would recommend the check (I added the check purposefully when adding it) to avoid unnecessary pip lookup.

This is not a big deal @balamurugana I think that extra check is useless Unixy style which we don't have to do. It also ensures that `make check` uses the latest as needed. 

I purposefully removed it.

balamurugana: I am puzzled how the PR https://github.com/minio/minio-py/pull/896 succeeded the workflow. May be a latest github workflow behaviour. But we need to fix the workflow script, not the Makefile.

harshavardhana: @balamurugana makefile should be fixed as well its the same as workflow. The checks are useless we are not saving anything by not hitting pip - we are trying to optimize things for limited benefits. 

In-fact its a problem because you have be using stale pylint,isorts on your system.

harshavardhana: If you see just upgrading fixed more things in the codebase @balamurugana - leaving things stale might incur costs that we not aware of. The Internet is cheap and we should always upgrade...

balamurugana: I do not agree this point because we have no control where the executables come from. There is a possibility to use older executables too. This is the reason the check prevents unnecessary install as well.

harshavardhana: I foresee more benefits @balamurugana, I prefer that we use the latest and fix bugs to move forward. I see no advantage with older binaries - there is really no demonstrable advantage which we need to hold on to. If the installation is already the latest it doesn't do anything and its rather quick.

So I will keep the PR as-is for now I am not planning to change it back to the older style.

PR #948, opened 2020-08-03
queue and queue_empty were removed from compat.py in minio-py 6.0

ephes:
But the imports in `examples/progress.py` were not updated -> import error. PR after fixing it in one of my projects after upgrading to 6.0.


kannappanr:
@ephes Thanks for the PR

PR #953, opened 2020-08-06
merge list_objects_v2() into list_objects() method

balamurugana:


harshavardhana:
> LGTM, although this looks like a breaking change, maybe we need to increase major version number now

Yes, unfortunately, we will have to do that @vadmeste

tests/unit/list_objects_test.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # MinIO Python Library for Amazon S3 Compatible Cloud Storage,
-# (C) 2015, 2016 MinIO, Inc.
+# (C) 2015 MinIO, Inc.
```
kannappanr, line 3:
```suggestion
# (C) 2015-2020 MinIO, Inc.
```

balamurugana:
Depends on #966

examples/list_objects.py
```diff
@@ -32,8 +32,8 @@
 
 # List all object paths in bucket that begin with my-prefixname using
 # V2 listing API.
-objects = client.list_objects_v2('my-bucketname', prefix='my-prefixname',
-                                 recursive=True)
+objects = client.list_objects('my-bucketname', prefix='my-prefixname',
```
vadmeste, line 35: This example has duplicated code after your change, I guess you need to use api v1 in the first line and update some wording

vadmeste (commented):
One comment, LGTM otherwise

examples/list_objects.py
balamurugana: done

vadmeste (approved):
LGTM & tested

PR #958, opened 2020-08-14
Fix creds check on chain provider

febg:


febg:
@balamurugana 
Maybe checking for empty string is not necessary... but checking for `creds.access_key` and `creds.secret_key` to have a value is crucial here.

As of now we are only checking if `creds` has a value.
```
if creds:
    ......
```
This fails because all providers return a `Value` object. 

This change was introduced in a refactor that was done some months ago. Before this refactor the chain provider was working as expected



So for example if you use a chain provider like this:
```client = Minio('s3.amazonaws.com',
               credentials=Credentials(
                   provider=Chain(
                        providers=[
                                EnvMinio()
                                EnvAWS(),
                                IAMProvider(),
                        ]
                   )
               )
           )

```
If `EnvMinio` credentials are not provided, the Chain provider will fail to try the other two providers.

balamurugana:
@febg Please fix those env providers to return `null` creds. That is the right fix. Anyway I will simplify `credentials` module further simple.

balamurugana:
Fixed addressing issue properly in https://github.com/minio/minio-py/pull/963

Issue #956, opened 2020-08-13
Unable to use AWS ECS task role credentials


NickLavrov:
IAM provider source code: https://github.com/minio/minio-py/blob/1ee647eaf9507e04b21404a251bc6953b9861c13/minio/credentials/providers.py#L187

The issue is that `creds_path = "/latest/meta-data/iam/security-credentials"` is hardcoded.

On an EC2 instance, you can run `curl http://169.254.169.2data/iam/security-credentials` to get the iam role, then `curl http://169.254.169.2data/iam/security-credentials/ROLENAME` to get the credentials.

For an ECS task running with a task iam role on an ec2 instance, you can get credentials if you run `curl 169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`.  The value for `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is typically like: `/v2/credentials/54da6c7a-84ff-41dc-a160-21b8a8bc7fb0`. The provider lets you set `endpoint`, but since `creds_path` is hardcoded, I cannot get the credentials.

Also, the logic for "Get credentials of first role." is not required in the ECS task method.

I suspect a fix would involve either making `creds_path` also configurable (and also the logic to get the first role, since it directly gives the credentials), or checking if `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is present in the environment variables and changing behavior accordingly.

Example results from `curl 169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`:

```
{"RoleArn":"arn:aws:iam::ACCOUNTID:role/ROLENAME","AccessKeyId":"ASIAXXXX","SecretAccessKey":"XXXXXX","Token":"TOKENTEXT","Expiration":"2020-08-13T06:11:55Z"}
```

balamurugana:
@NickLavrov 
> For an ECS task running with a task iam role on an ec2 instance, you can get credentials if you run `curl 169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`. The value for `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is typically like: `/v2/credentials/54da6c7a-84ff-41dc-a160-21b8a8bc7fb0`. The provider lets you set `endpoint`, but since `creds_path` is hardcoded, I cannot get the credentials.

[...]

PR #960, opened 2020-08-17
IAMProvider accepts ECS IAM Task roles

NickLavrov:
Per this issue https://github.com/minio/minio-py/issues/956

This PR adds the `is_ecs_task` arg to `IAMProvider`. When it is set to `True`, it follows a slightly different URL path to get credentials.

If you run a task on AWS ECS without a given task role, then `curl http://169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` from within the container return a `404` error, and this would be caught by the error handling already in this class.

This would also require updates to the documentation on `IAMProvider`.

### DIFF
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -186,7 +186,7 @@ class FileMinioClient(Provider):
 
 class IAMProvider(Provider):
     """
-        IAM EC2 credential provider.
+        IAM EC2/ECS credential provider.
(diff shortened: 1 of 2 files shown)


balamurugana (approved):
We can take this PR as is. I am working on adding WebIdentity provider and modifying IAM provider to support to complete.

Issue #957, opened 2020-08-13
Mix of timezone-aware and naive datetime objects with AssumeRoleProvider


wilko77:
I was trying to update our code base to use minio 6.0.0 and ran into the following problem:

```
self = <minio.credentials.credentials.Credentials object at 0x110fad610>
force = False

    def get(self, force=False):
        """Get credentials from provider if needed."""
        if (
                force or
                not self._value or
>               (self._expiry and self._expiry < datetime.utcnow())
        ):
E       TypeError: can't compare offset-naive and offset-aware datetimes
```

I could track it down to the following code:
```
credentials_provider = AssumeRoleProvider( lambda: some_minio_client.get_assume_role_creds(restricted_upload_policy))
temp_credentials = Credentials(provider=credentials_provider)

restricted_mc_client = Minio(upload_endpoint, credentials=temp_credentials, region='us-east-1', secure=False)
```

The problem seems to be that `get_assume_role_creds` calls `parse_assume_role` which returns the expiry as a timezone-aware datetime object https://github.com/minio/minio-py/blob/e1d58b2599fed979fa8459cd6a22f1699dd919a6/minio/parsers.py#L531
However, in the `Credentials` class in the `get` method there is a test against a naive datetime object: https://github.com/minio/minio-py/blob/e1d58b2599fed979fa8459cd6a22f1699dd919a6/minio/credentials/credentials.py#L71

Am I using the `AssumeRoleProvider` incorrectly or is there a bug?

balamurugana:
Fixed by https://github.com/minio/minio-py/pull/963

PR #963, opened 2020-08-20
simplify credentials module

### LINKED ISSUE #956: Unable to use AWS ECS task role credentials

balamurugana:
* Added LDAP, ClientGrants and WebIdentity providers.
* Refactor IAM provider and renamed to IamAwsProvider.
* Refactor AssumeRole provider.
* Renamed FileAWSCredentials provider to AWSConfigProvider.
* Renamed Chain provider to ChainedProvider.
* Renamed EnvAWS provider to EnvAWSProvider.
* Renamed EnvMinio provider to EnvMinioProvider.
* Renamed FileMinioClient provider to MinioClientConfigProvider.
* Renamed Static provider to StaticProvider.

Fixes #956 #957


minio/credentials/providers.py
```diff
@@ -17,70 +17,181 @@
...
-        self._expiry = None
+    def __init__(
+            self, sts_endpoint, access_key, secret_key, duration_seconds=0,
+            policy=None, region=None, role_arn=None, role_session_name=None,
+            external_id=None, http_client=None,
+    ):
+        self._sts_endpoint = sts_endpoint
+        self._access_key = access_key
+        self._secret_key = secret_key
+        self._region = region
```
Praveenrajmani, line 98: shouldn't we fallback to "us-east-1" as default if `region` is not specified?

Praveenrajmani (approved):
Just few questions/doubts. Looks good to me otherwise

minio/api.py
```diff
@@ -154,22 +153,13 @@ def __init__(self, endpoint, access_key=None,
...
-            provider=Chain(
-                providers=[
-                    Static(access_key, secret_key, session_token),
-                    EnvAWS(),
-                    EnvMinio(),
-                ]
-            )
-        )
+        if access_key:
+            credentials = StaticProvider(access_key, secret_key, session_token)
```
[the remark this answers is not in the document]
balamurugana: Yes. This PR breaks backward compatibility to some extends. Previously getting credentials from env is wrong i.e. there is no way to create `Minio` with anonymous access unless `Chain` provider returns `None` credentials. This looks odd and no control to users.

As credentials is concerned, the user should have more control over by passing necessary credential provider. Additionally when access/secret keys are provided along with credential provider, its better to ignore credential provider and use access/secret keys (as static credentials). This is what I changed.

Note that Chain provider are meant to give valid access/secret keys, they are not credential fail-over mechanism i.e. if one access/secret keys are invalid, then fetch another access/secret from next provider.

minio/credentials/providers.py
balamurugana: I see `region` is optional in `minio-go`; however for signature v4 region must be needed. I will check why `region` is optional in `minio-go` and fix accordingly here.

balamurugana: I got clarified on this. We would need to use empty string as region if not provided. Fixing accordingly.

minio/api.py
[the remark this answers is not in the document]
Praveenrajmani: understood. Thanks!

Praveenrajmani (approved):
LGTM

nitisht (approved):

PR #967, opened 2020-08-24
replace {list,remove}_incomplete_uploads with private methods.

balamurugana:
As `minio-py` does not leave stale multipart uploads,
`list_incomplete_uploads()` and `remove_incomplete_uploads()` methods are
removed, but introduced new private methods `_list_mulitpart_uploads()` and
`_list_parts()` for low level usage.

PR #968, opened 2020-08-28
Move http execution at one place

balamurugana:
Errors and exceptions:
* `ValueError` is raised any invalid arguments instead of specific exceptions.
* `S3Error` is raised instead of individual S3 error code specific exceptions.
* `MinioException` is a base exception and `MinioError` is removed.

Signer:
* Signature V4 calculation is simplified.
* Multi-spaced header values are properly handled now.
* Multi-valued headers/query parameters are properly handled now.
* sign_v4(), presign_v4() and post_presign_v4() do not change any headers
  values; and sign_v4() adds Authorization header only.

XML parser:
* Time values are kept as UTC instead of local time.

APIs:
* HTTP execution is done only in _url_open() via _execute().
* HTTP traces happen only in _url_open().
* Bucket operation calls do not send trailing '/' in URI.
* Option to enable/disable accelerate/dualstack/virtual-style for
  Amazon S3 endpoints and no specific region handling is done.
* fget_object() has an optional argument to temporary file to stage
  object download.
* put_object() supports unknown sized stream and optional argument to control
  parallel uploads. Now progress object is updated more precisely.
* listen_bucket_notification() does not leak connections now.


balamurugana:
Depends on https://github.com/minio/minio-py/pull/953 https://github.com/minio/minio-py/pull/963 https://github.com/minio/minio-py/pull/967

ebozduman (approved):
LGTM

Praveenrajmani (approved):
LGTM

PR #971, opened 2020-09-02
Remove unnecessary requirement on "future"

brianhelba:
`future` was only required for Python 2, which is not supported by minio.


balamurugana (approved):

ebozduman (approved):
LGTM

PR #998, opened 2020-10-14
Fix query params type

dtaniwaki:
I got the following error when I tried `IamAwsProvider` because there's a typo.

```
>>> from minio.credentials import IamAwsProvider
>>>
>>> cred = IamAwsProvider()
>>> cred.retrieve()
  File "<stdin>", line 1, in <module>
  File "/usr/local/src/minio/minio/credentials/providers.py", line 432, in retrieve
    self._credentials = provider.retrieve()
  File "/usr/local/src/minio/minio/credentials/providers.py", line 576, in retrieve
    query_params["Action"] = "AssumeRoleWithWebIdentity"
TypeError: 'set' object does not support item assignment
```

### DIFF
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -563,7 +563,7 @@ class WebIdentityClientGrantsProvider(Provider):
 
         jwt = self._jwt_provider_func()
 
-        query_params = {"Version", "2011-06-15"}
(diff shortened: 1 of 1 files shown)


balamurugana:
Fixed by https://github.com/minio/minio-py/pull/1000

PR #999, opened 2020-10-14
Allow to pass region to IamAwsProvider

dtaniwaki:
I use a different environment variable from `AWS_REGION` and the Minio client accepts it because it receives the value via the arguments. I think we can allow it to `IamAwsProvider` too.

### DIFF
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -363,7 +363,7 @@ def _get_jwt_token(token_file):
 class IamAwsProvider(Provider):
     """Credential provider using IAM roles for Amazon EC2/ECS."""
 
-    def __init__(self, custom_endpoint=None, http_client=None):
+    def __init__(self, custom_endpoint=None, http_client=None, region=None):
         self._custom_endpoint = custom_endpoint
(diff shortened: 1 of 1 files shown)


balamurugana (changes requested):
This is not the way IAM roles work in Amazon EC2/ECS environment. In the environment, all necessary ENV VARS are populated.

dtaniwaki:
Should I just set `custom_endpoint` with the region?

dtaniwaki:
`AWS_REGION` is not set in my EKS automatically.

balamurugana:
@dtaniwaki 

> `AWS_REGION` is not set in my EKS automatically.

This means you don't need region to access your STS service. I don't see this PR useful.

dtaniwaki:
Ah, I understand what happened.

https://github.com/minio/minio-py/blob/316c3243f5393236667a48a337481db1ba9c6f0d/minio/credentials/providers.py#L421-L423

The format function prints `None` if it's None and makes the URL invalid.

dtaniwaki:
Made another PR https://github.com/minio/minio-py/pull/1001 and closing this one.

PR #1003, opened 2020-10-15
Add python 3.9 support in github workflow.

menishmueli:
- Add python 3.9 to tested python versions
- Use  setup-python@v2 (v1 has bug, see https://github.com/actions/setup-python/issues/154)
- Update setup.py that the library supports python 3.9

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,8 +72,6 @@ setup(
         'Natural Language :: English',
         'Operating System :: OS Independent',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 3.4',
-        'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -76,6 +76,8 @@
         'Programming Language :: Python :: 3.5',
```
balamurugana, line 76: Please remove 3.4 and 3.5

menishmueli: The readme states that python 3.4 is supported, so if any the CI should also contain 3.4 and 3.5, right?
Or the readme is wrong and it works only on 3.6+?

balamurugana: Please fix README.md as well. We do not support EOL'ed versions.

menishmueli: done

balamurugana (approved):

kannappanr (approved):

PR #1018, opened 2020-11-11
Simplify select_object_content() response stream

balamurugana:


vadmeste:
@balamurugana, could you test this example again examples/select_object_content.py ?
1. You need to open the output file in binary mode
2. It doesn't look that "select * from s3object" returned all records as per my testing.

balamurugana:
@vadmeste 

> @balamurugana, could you test this example again examples/select_object_content.py ?
> 
>     1. You need to open the output file in binary mode
> 
>     2. It doesn't look that "select * from s3object" returned all records as per my testing.

1. I refactor the example completely.
2. There is a [functional test](https://github.com/balamurugana/minio-py/blob/Simplify-select_object_content-response-stream/tests/functional/tests.py#L257) doing the same. As the test passes, I don't think there is an issue. Can you provide me failures and debugging information?

sinhaashish (dismissed):

sinhaashish (approved):

vadmeste (approved):
LGTM & tested

PR #1021, opened 2020-11-14
Add compose_object() and enhance copy_object() APIs

balamurugana:

PR #1022, opened 2020-11-19
Enhance put object api to support more parameters

balamurugana:
Depends on https://github.com/minio/minio-py/pull/1021


examples/put_object.py
```diff
@@ -14,43 +14,121 @@
...
+
+client = Minio(
+    "s3.amazonaws.com",
+    access_key="YOUR-ACCESSKEYID",
+    secret_key="YOUR-SECRETACCESSKEY",
+)
+
+# Upload data.
+result = client.put_object(
+    "my-bucket", "my-object", io.StringIO("hello"), 5,
```
vadmeste, line 34: It looks like this needs to be io.BytesIO(b"hello")

examples/put_object.py
```diff
@@ -14,43 +14,121 @@
...
+)
+print(
+    "created {0} object; etag: {1}, version-id: {2}".format(
+        result.object_name, result.etag, result.version_id,
+    ),
+)
+
+# Upload unknown sized data.
+result = client.put_object(
+    "my-bucket", "my-object", data, size=-1, part_size=10*1024*1024,
```
vadmeste, line 44: `size` is not defined as keyword

vadmeste (commented):
Some comments

examples/put_object.py
balamurugana: Done

examples/put_object.py
balamurugana: Done

harshavardhana (approved):

kannappanr (approved):
LGTM

PR #1025, opened 2020-11-20
Make code property check optional in IamAwsProvider

balamurugana:


vadmeste:
@balamurugana what does this fix exactly ? Is there any user experiencing issues ?

balamurugana:
@vadmeste Please refer issue https://github.com/minio/minio-java/issues/1108

minio/credentials/providers.py
```diff
@@ -387,10 +382,10 @@ def fetch(self, url):
 
         res = _urlopen(self._http_client, "GET", url)
         data = json.loads(res.data)
-        if data["Code"] != "Success":
+        if data.get("Code", "Success") != "Success":
```
vadmeste, line 385: Maybe let's check for   "AccessKeyId", "SecretAccessKey" and "Token" fields, if they are non empty then we can assume this is successful call

vadmeste (commented):
One comment

minio/credentials/providers.py
balamurugana: If those fields don't have proper values, it will fail at line no 393. I think this check is good enough to address other cases. However this fix is only testable in AWS ECS env.

vadmeste (approved):
LGTM

krisis (approved):

PR #1045, opened 2020-12-09
fix: bring back previous change to functional test script

harshavardhana:


run_functional_tests.sh
```diff
@@ -34,9 +34,16 @@ function run_minio_server() {
     tests/functional/minio server --config-dir tests/functional/.cfg tests/functional/.d{1...4} >tests/functional/minio.log 2>&1 &
 }
 
-run_minio_server
-sleep 3
+if [ -z ${SERVER_ENDPOINT+x} ]; then
+    run_minio_server
+    trap 'pkill -9 minio' INT
```
balamurugana, line 39: this will kill any minio process owned by the user running in the system. you would need to kill only if the process the script created. you could redirect failure to `/dev/null` to avoid error message.

run_functional_tests.sh
```diff
@@ -34,9 +34,16 @@ function run_minio_server() {
...
-SECRET_KEY=minio123 ACCESS_KEY=minio SERVER_ENDPOINT=localhost:9000 ENABLE_HTTPS=0 MINT_MODE=full PYTHONPATH=$PWD python tests/functional/tests.py
+    export MINT_MODE=full
+    export SERVER_ENDPOINT=localhost:9000
+    export ACCESS_KEY=minio
+    export SECRET_KEY=minio123
+    export ENABLE_HTTPS=0
+fi
+
+PYTHONPATH=$PWD python tests/functional/tests.py
 pkill -9 minio
```
balamurugana, line 49: same as above

run_functional_tests.sh
```diff
@@ -34,9 +34,16 @@ function run_minio_server() {
     tests/functional/minio server --config-dir tests/functional/.cfg tests/functional/.d{1...4} >tests/functional/minio.log 2>&1 &
 }
 
-run_minio_server
-sleep 3
+if [ -z ${SERVER_ENDPOINT+x} ]; then
```
balamurugana, line 37: I don't think we need to remove `minio` binary if already downloaded. The right way to do is to download it if not found.

balamurugana (changes requested):

run_functional_tests.sh
harshavardhana: The point Is to always download fresh binary to, since you don't want to be running tests against some older binary locally.

run_functional_tests.sh
harshavardhana: Yeah will bring it back..

balamurugana (approved):

BigUstad (approved):
LGTM

PR #1046, opened 2020-12-09
Remove configparser as dependency

### LINKED ISSUE #983: configparser is part of the interpreter in Python 3
This project specifies that configparser is required in setup.py: https://github.com/minio/minio-py/blob/0ca673bd886e7f63bdddff814b1aacb1d16a2adc/setup.py#L45

According to the Python 3 documentation (https://docs.python.org/3/library/configparser.html) and the configparser documentation (https://github.com/jaraco/configparser/blob/master/README.rst), configparser is part of the interpreter in Python 3, unless it is imported as `from backports import configparser`, which this project does not do. This requirement needs a Python interpreter specification on it, or it can simply be removed since Python 2 support has been removed in #973.

balamurugana:
Fixes #983

### DIFF
--- a/.github/workflows/pythonpackage-linux.yml
+++ b/.github/workflows/pythonpackage-linux.yml
@@ -27,7 +27,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip setuptools
-        pip install certifi configparser urllib3 mock nose
+        pip install certifi urllib3 mock nose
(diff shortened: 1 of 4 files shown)


ebozduman (approved):
LGTM

vadmeste (approved):
LGTM, tested minio_with_aws_config_provider.py example without configparser installed as a module

PR #1053, opened 2021-01-05
Allow empty prefix in Filter

balamurugana:


harshavardhana (changes requested):
One more change needed is to treat `element.text is None` as `""` empty string.

harshavardhana (approved):

Praveenrajmani (approved):

PR #1063, opened 2021-01-16
fix docstring and error handling in _validate_size()

mykola-mokhnach:
I've just decided to start with something simple to get better acquainted with the codebase. The PR simply addresses the warnings that PyCharm linter is complaining about

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -2603,7 +2603,8 @@ class Minio:  # pylint: disable=too-many-public-methods
         Example::
             client.delete_object_lock_config("my-bucket")
         """
-        self.set_object_lock_config(bucket_name, ObjectLockConfig(None, None, None))
+        lock_config = ObjectLockConfig(None, None, None)
+        self.set_object_lock_config(bucket_name, lock_config)
 
     def get_object_lock_config(self, bucket_name):
         """
--- a/minio/api.py
+++ b/minio/api.py
@@ -2603,8 +2603,9 @@ class Minio:  # pylint: disable=too-many-public-methods
         Example::
             client.delete_object_lock_config("my-bucket")
         """
-        lock_config = ObjectLockConfig(None, None, None)
-        self.set_object_lock_config(bucket_name, lock_config)
+        self.set_object_lock_config(
+            bucket_name, ObjectLockConfig(None, None, None)
+        )
 
     def get_object_lock_config(self, bucket_name):
         """
--- a/minio/api.py
+++ b/minio/api.py
@@ -2603,9 +2603,7 @@ class Minio:  # pylint: disable=too-many-public-methods
         Example::
             client.delete_object_lock_config("my-bucket")
         """
-        self.set_object_lock_config(
-            bucket_name, ObjectLockConfig(None, None, None),
-        )
+        self.set_object_lock_config(bucket_name, ObjectLockConfig(None, None, None))
 
     def get_object_lock_config(self, bucket_name):
         """
--- a/minio/api.py
+++ b/minio/api.py
@@ -510,7 +510,7 @@ class Minio:  # pylint: disable=too-many-public-methods
         """
         Enable http trace.
 
-        :param output_stream: Stream for writing HTTP call tracing.
+        :param stream: Stream for writing HTTP call tracing.
         """
         if not stream:
             raise ValueError('Input stream for trace output is invalid.')
@@ -755,7 +755,7 @@ class Minio:  # pylint: disable=too-many-public-methods
         Set notification configuration of a bucket.
 
         :param bucket_name: Name of the bucket.
-        :param :class:`NotificationConfig <NotificationConfig>` object.
+        :param config: class:`NotificationConfig <NotificationConfig>` object.
 
         Example::
             config = NotificationConfig(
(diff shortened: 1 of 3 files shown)


minio/commonconfig.py
```diff
@@ -417,18 +417,6 @@ def __init__(self, bucket_name, object_name, region=None, version_id=None,
 
     def _validate_size(self, object_size):
         """Validate object size with offset and length."""
-        def _get_error(name, value):
```
balamurugana, line 420: Use this function below in all `raise ValueError()`

minio/api.py
```diff
@@ -2602,7 +2604,7 @@ def delete_object_lock_config(self, bucket_name):
             client.delete_object_lock_config("my-bucket")
         """
         self.set_object_lock_config(
-            self, bucket_name, ObjectLockConfig(None, None, None),
+            bucket_name, ObjectLockConfig(None, None, None),
```
balamurugana, line 2607: Now you could keep it one liner if possible.

balamurugana (commented):
This project uses pylint and autopep8. However your changes are welcome.

minio/commonconfig.py
mykola-mokhnach: could you please provide an example of how it should be used?

minio/api.py
mykola-mokhnach: changed

mykola-mokhnach:
It would be interesting to know the purpose of unused `element` argument in most of the `toxml` calls, for example https://github.com/minio/minio-py/blob/cae7a58ecef451ce390478b76fc54927ae8bc6bd/minio/sseconfig.py#L94
Is it just needed to unify the method signature?

minio/api.py
```diff
@@ -2601,9 +2603,8 @@ def delete_object_lock_config(self, bucket_name):
         Example::
             client.delete_object_lock_config("my-bucket")
         """
-        self.set_object_lock_config(
-            self, bucket_name, ObjectLockConfig(None, None, None),
-        )
+        lock_config = ObjectLockConfig(None, None, None)
+        self.set_object_lock_config(bucket_name, lock_config)
```
balamurugana, line 2607: As you are unable to keep it one liner, just remove extra `self`.

minio/commonconfig.py
balamurugana: `raise ValueError(_get_error("offset", self._offset))`

balamurugana:
> It would be interesting to know the purpose of unused `element` argument in most of the `toxml` calls, for example
> 
> https://github.com/minio/minio-py/blob/cae7a58ecef451ce390478b76fc54927ae8bc6bd/minio/sseconfig.py#L94
> 
> Is it just needed to unify the method signature?

I don't see unused `element` argument. Class denote root element doesn't use passed `element` parameter on purpose.

minio/api.py
mykola-mokhnach: nope, the linter complains about the line length (over 80 chars) in such case

minio/commonconfig.py
mykola-mokhnach: changed

balamurugana: Please do required changes than changing function name and it's behaviour.

minio/api.py
balamurugana: I am saying remove `self` usage in original code. There is no need to have `lock_config` variable.

minio/commonconfig.py
mykola-mokhnach: I did change the function, because I see an unnecessary repetition of ValueError. Also, I don't see a need to add underscore to the function name, because it's internal and is not visible anywhere outside of this particular method scope

minio/api.py
mykola-mokhnach: I did and got these linter errors: https://github.com/minio/minio-py/runs/1714888958

Shall I change linter settings and increase the maximum length of a single line?

balamurugana: I am saying below
```py
        self.set_object_lock_config(
            bucket_name, ObjectLockConfig(None, None, None),
        )
```

mykola-mokhnach: ok, reverted it back

balamurugana (approved):

minio/commonconfig.py
```diff
@@ -417,9 +417,9 @@ def __init__(self, bucket_name, object_name, region=None, version_id=None,
...
+        def make_error(name, value):
```
ebozduman, line 420: This is what I would prefer and implement, and you are welcome to ignore this comment if you'd like;
`make_error` sounds like there is no error and you are making it in my "opinion".
We hit the error and we are forming/putting together a value error message to be raised. So, I would 'personally' call `make_error` method => `raise_value_error` and raise the `ValueError` inside the method, instead of returning the `ValueError` and then raise the returned value.

As I said, it is up to you, so no matter what it is a LGTM.

mykola-mokhnach: I thought about it, but unfortunately there is no good way in Python to hint to the interpreter that the particular method is always going to throw an exception. Thus linters cannot properly predict the code flow after such method is called, because they cannot guess the correct flow. Example


```

def method_that_always_raises():
    raise ValueError


def some_other_method():
        method_that_always_raises()
        # here linters are lost and assume the second operator would still be executed
        assert False, 'This never happens'


def method_that_returns_exc():
    return ValueError


def some_other_method2():
        raise method_that_returns_exc()
        # here linters know for sure that method execution would never reach the next line
        assert False, 'This never happens'
```

ebozduman (approved):
LGTM

minio/commonconfig.py
ebozduman: I see.
Thanks.

PR #1067, opened 2021-01-19
[WIP] refactor: Always use HTTPHeaderDict where possible while working with HTTP headers

mykola-mokhnach:
Plain dict should not be used for such purpose, because there is a mess with header names, which are case-insensitive by nature. Assigning header values with different case (e.g. 'Content-Length' vs 'content-length') might create serious issues, which would be hard to track

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -34,7 +34,6 @@ from xml.etree import ElementTree as ET
 
 import certifi
 import urllib3
-from urllib3._collections import HTTPHeaderDict
 
 from . import __title__, __version__, time
 from .commonconfig import COPY, REPLACE, ComposeSource, CopySource, Tags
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
It seems like the logic behind headers is complicated, so I can hardly figure out what exactly needs to be fixed in order to make tests passing. In general, header values must only be strings, but I can observe that integers, lists and tuples are being assigned to dictionary values instead. And then converted to plain HTTP headers right before being passed to urllib3 call. 
This means such refactoring might touch much more places in the code and/or even change the architecture, which would be too much for a small PR :(

@balamurugana I wonder if it would be possible to split such refactoring into smaller parts and if this would be possible in general without breaking anything important

balamurugana:
@mykola-mokhnach IMO this change is unnecessary and I don't see any issues currently. If you find any bugs specifically, please send the fix.

PR #1089, opened 2021-03-02
Fix wrong key name in retrieve

vonera:
retrieve() checks for the key "hosts" but in the current implementation it is called "aliases"


vadmeste:
@vonera, thanks for the PR. 

Can you check 'hosts' as a fallback ? because many mc deployments still have the old format

PR #1090, opened 2021-03-02
Add new config format support in MinioClientConfigProvider

vonera:
Keep both the old format and the new format key to support older and the current mc config files

### DIFF
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -307,17 +307,14 @@ class MinioClientConfigProvider(Provider):
         try:
             with open(self._filename) as conf_file:
                 config = json.load(conf_file)
-            if "aliases" in config:
-                key = "aliases"
-            elif "hosts" in config:
-                key = "hosts"
-            else:
+            aliases = config.get("hosts") or config.get("aliases")
+            if not aliases:
                 raise ValueError(
                     "invalid configuration in file {0}".format(
                         self._filename,
                     ),
                 )
-            creds = config.get(key).get(self._alias)
+            creds = aliases.get(self._alias)
             if not creds:
                 raise ValueError(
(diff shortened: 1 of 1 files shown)


minio/credentials/providers.py
```diff
@@ -307,13 +307,17 @@ def retrieve(self):
...
+            elif "hosts" in config:
+                key = "hosts"
+            else:
                 raise ValueError(
                     "invalid configuration in file {0}".format(
                         self._filename,
                     ),
                 )
-            creds = config.get("hosts").get(self._alias)
+            creds = config.get(key).get(self._alias)
```
balamurugana, line 320: Use more pythonic like below
```py
diff --git a/minio/credentials/providers.py b/minio/credentials/providers.py
index 1447a59..7765e38 100644
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -307,13 +307,14 @@ class MinioClientConfigProvider(Provider):
         try:
             with open(self._filename) as conf_file:
                 config = json.load(conf_file)
-            if not config.get("hosts"):
+            aliases = config.get("hosts") or config.get("aliases")
+            if not aliases:
                 raise ValueError(
                     "invalid configuration in file {0}".format(
                         self._filename,
                     ),
                 )
-            creds = config.get("hosts").get(self._alias)
+            creds = aliases.get(self._alias)
             if not creds:
                 raise ValueError(
                     (
```

balamurugana (changes requested):

vonera (commented):
changed as suggested

balamurugana (approved):

balamurugana (approved):

kannappanr (approved):

PR #1099, opened 2021-03-23
Handle empty stdout properly in MinioAdmin

cbows:
`json.loads(line)` would throw an error if line is empty (e.g. `line == b''`).
To fix it, simply check if line is truthy (i.e. not empty).

Also, since `proc.stdout` is of type `bytes` the argument passed to
`proc.stdout.split` was changed to `b"\n"`.

### DIFF
--- a/minio/minioadmin.py
+++ b/minio/minioadmin.py
@@ -48,9 +48,12 @@ class MinioAdmin:
             capture_output=True,
             timeout=self._timeout,
             check=True,
+            text=True,
         )
+        if not proc.stdout:
+            return [] if multiline else {}
         if multiline:
-            return [json.loads(line) for line in proc.stdout.split("\n")]
+            return [json.loads(line) for line in proc.stdout.splitlines()]
         return json.loads(proc.stdout)
 
     def service_restart(self):


balamurugana:
@cbows 
> json.loads(line) would throw an error if line is empty (e.g. line == b'').
> To fix it, simply check if line is truthy (i.e. not empty).

On which admin API do you get this error?

cbows:
I got it on `policy_list`, which had an empty `b''` as the last item in the list.

minio/minioadmin.py
```diff
@@ -50,7 +50,7 @@ def _run(self, args, multiline=False):
             check=True,
```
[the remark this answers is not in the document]
cbows: Done

minio/minioadmin.py
```diff
@@ -48,9 +48,10 @@ def _run(self, args, multiline=False):
             capture_output=True,
             timeout=self._timeout,
             check=True,
+            text=True,
         )
```
balamurugana, line 52: Add a fix generically to avoid any error on `json.loads()`  here
```py
if not proc.stdout:
    return [] if multiline else {}
```

balamurugana (changes requested):

minio/minioadmin.py
cbows: Ok

minio/minioadmin.py
```diff
@@ -48,9 +48,12 @@ def _run(self, args, multiline=False):
...
+        if not proc.stdout:
+            return [] if multiline else {}
         if multiline:
-            return [json.loads(line) for line in proc.stdout.split("\n")]
+            return [json.loads(line) for line in proc.stdout.split("\n") if line]
```
balamurugana, line 56: No change required in this line.

balamurugana (changes requested):

minio/minioadmin.py
cbows: The error I had actually occurs when proc.stdout is not empty, but has an empty `b""` after `split`.

cbows: Here is an example:

The response:
```
'{"status":"success","policy":"diagnostics","isGroup":false}\n{"status":"success","policy":"readonly","isGroup":false}\n{"status":"success","policy":"readwrite","isGroup":false}\n{"status":"success","policy":"writeonly","isGroup":false}\n'
```
becomes
```python
['{"status":"success","policy":"diagnostics","isGroup":false}',
 '{"status":"success","policy":"readonly","isGroup":false}',
 '{"status":"success","policy":"readwrite","isGroup":false}',
 '{"status":"success","policy":"writeonly","isGroup":false}',
 '']
```
which fails at the very end because the trailing new-line results in empty string.

cbows: you could also do something like `proc.stdout.strip().split("\n")` to get rid of any leading/trailing whitespace.

balamurugana: Use `.splitlines()` instead of `.split("\n")`

cbows: That still doesn't handle cases where response is `"a\n\nb\nc\n"`, but it's good enough for me.

balamurugana (approved):

harshavardhana (approved):

PR #1103, opened 2021-03-27
make_bucket: use region passed via constructor

### LINKED ISSUE #1102: make_bucket should use base_url's region as location if configured
I ran into an issue where if I configure a region when creating a client like so:

```
 client = Minio(
        settings.MINIO_SERVER,
        access_key=settings.MINIO_ACCESS_KEY,
        secret_key=settings.MINIO_SECRET_KEY,
        secure=settings.MINIO_USE_HTTPS,
        region='my-region',
    )
```

by default

`client.make_bucket('bucket_name')` attempts to create a bucket in `us-east-1`, which fails with a `AuthorizationHeaderMalformed` exception

I would expect that https://github.com/minio/minio-py/blob/ac047e7b11610239077599859535b5dab39a01f5/minio/api.py#L622 should be changed so that it defaults to `self._base_url.region` if set

balamurugana:
Fixes #1102

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -619,7 +619,7 @@ class Minio:  # pylint: disable=too-many-public-methods
                         self._base_url.region, location,
                     ),
                 )
-        location = location or "us-east-1"
+        location = self._base_url.region or location or "us-east-1"
         headers = (
             {"x-amz-bucket-object-lock-enabled": "true"}
             if object_lock else None


minio/api.py
```diff
@@ -619,7 +619,7 @@ def make_bucket(self, bucket_name, location=None, object_lock=False):
                         self._base_url.region, location,
                     ),
                 )
-        location = location or "us-east-1"
+        location = self._base_url.region or location or "us-east-1"
```
kannappanr, line 622:
```suggestion
        location = location or self._base_url.region or "us-east-1"
```

kannappanr: Should it be this, as the user is explicitly specifying the location they want to use, even if the region is set in the constructor?

balamurugana: No. It is not allowed. It will make confusion in other APIs.

ebozduman: This is a general comment.
Wouldn't it be better to define a constant for the default region instead of hard coding it in several different places (5) in the code?
```
_DEFAULT_REGION = "us-east-1"
```

ebozduman (changes requested):

minio/api.py
balamurugana: I think this is out of scope of this PR.

kannappanr: so, if `region` is specified in the constructor, and also specified in the `make_bucket` call and they are different, do we reject it?

balamurugana: yep

ebozduman (approved):
LGTM

kannappanr (approved):
LGTM

PR #1111, opened 2021-05-24
fix: pylint build issues in minio-py

harshavardhana:
Bonus: also stringify the `key` for
additional metadata, otherwise key's with
numeric names might fail with key ValueError.

### DIFF
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -127,8 +127,7 @@ def _get_part_info(object_size, part_size):
         return part_size, -1
 
     if part_size > 0:
-        if part_size > object_size:
-            part_size = object_size
+        part_size = min(part_size, object_size)
         return part_size, math.ceil(object_size / part_size) if part_size else 1
 
     part_size = math.ceil(
@@ -322,7 +321,7 @@ def _metadata_to_headers(metadata):
 
 def normalize_headers(headers):
     """Normalize headers by prefixing 'X-Amz-Meta-' for user metadata."""
-    headers = headers.copy() if headers else {}
+    headers = {str(key): value for key, value in (headers or {}).items()}
 
     def guess_user_metadata(key):
         key = key.lower()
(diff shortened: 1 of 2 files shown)


minio/api.py
```diff
@@ -159,9 +159,6 @@ def __init__(self, endpoint, access_key=None,
             )
         )
 
-    def __del__(self):
```
balamurugana, line 162: This was added by https://github.com/minio/minio-py/pull/1088. Removal might break

minio/signer.py
```diff
@@ -59,7 +59,7 @@ def _get_canonical_headers(headers):
 
     canonical_headers = {}
     for key, values in headers.items():
-        key = key.lower()
+        key = str(key).lower()
```
balamurugana, line 62: As you have converted non-string to string via genheaders() -> normalize_headers() -> _metadata_to_headers(), there is no need to do it again.

Makefile
```diff
@@ -7,7 +7,7 @@ getdeps:
 
 check: getdeps
 	@echo "Running checks"
-	@pylint --reports=no --score=no --disable=R0401,R0801 minio/*py
+	@pylint --reports=no --score=no --disable=R0401,R0801,R1732 minio/*py
```
balamurugana, line 10: Instead of disabling here, disable checks specific to the code like
```py
        self._sem.acquire() # pylint: disable=consider-using-with
```

balamurugana (dismissed):

minio/api.py
harshavardhana: Yes on purpose @balamurugana this doesn't build anymore with minio-py master and latest `pylint`

minio/signer.py
harshavardhana: This is to specifically fix a scenario when `kv = dict(); kv['1'] = 'value'`  - I saw that this is not supported @balamurugana

kannappanr (dismissed):
LGTM

balamurugana (approved):

PR #1118, opened 2021-06-24
listen_bucket_notification(): return EventIterable object as controlled iterator.

### LINKED ISSUE #1117: How to properly timeout a notification listener?
I need to load an object from MinIO as soon as it is uploaded. For that, I use the MinIO python client's `Minio.listen_bucket_notification()` and listen for the "s3:ObjectCreated:*" event. The client only listens for a single object.

It can happen that the file does not get created in time. So I want to stop listening for this notification after a while. Looking at the implementation (https://github.com/minio/minio-py/blob/master/minio/api.py#L868), I don't see an option for a timeout, or any proper way for me to close the response and release the connection to MinIO from the outside.

What is the proper way to time out a MinIO client's `listen_bucket_notification()` iterator?

balamurugana:
Fixes #1117

Signed-off-by: Bala.FA <bala.gluster@gmail.com>

### DIFF
--- a/minio/datatypes.py
+++ b/minio/datatypes.py
@@ -33,6 +33,11 @@ from .signer import get_credential_string, post_presign_v4
 from .time import from_iso8601utc, to_amz_date, to_iso8601utc
 from .xml import find, findall, findtext
 
+try:
+    from json.decoder import JSONDecodeError
+except ImportError:
+    JSONDecodeError = ValueError
+
 
 class Bucket:
     """Bucket information."""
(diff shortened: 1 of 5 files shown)


minio/datatypes.py
```diff
@@ -755,3 +760,49 @@ def parse_copy_object(response):
...
+            self._response.release_conn()
+            self._response = None
+
+    def __iter__(self):
+        return self
+
+    def _get_records(self):
+        """Get event records from response stream."""
+        try:
+            line = self._stream.next().strip()
```
neuneck, line 786: `self._stream` is a generator, which does not have a `next()` method. Instead using
```
line = next(self._stream).strip()
```
works.

neuneck (approved):

minio/datatypes.py
balamurugana: Thanks for testing the PR. I just modified to use `__next__()`. Hope this works fine.

neuneck: Yep, it still works.

neuneck (approved):

paddySchuch (approved):

harshavardhana (approved):

PR #1122, opened 2021-07-02
Fixed a rare issue where a RuntimeError would be raised when closing …

neuneck:
…an EventIterable.

### DIFF
--- a/minio/datatypes.py
+++ b/minio/datatypes.py
@@ -815,6 +815,9 @@ class EventIterable:
 
     def _close_response(self):
         """Close response."""
+        if self._stream:
+            self._stream.close()
+            self._stream = None
         if self._response:
             self._response.close()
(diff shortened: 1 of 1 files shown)


minio/datatypes.py
```diff
@@ -815,6 +815,9 @@ def __init__(self, func):
 
     def _close_response(self):
         """Close response."""
+        if self._stream:
+            self._stream.close()
+            self._stream = None
         if self._response:
```
balamurugana, line 821: This closes underneath stream as well. I would need to check your `urllib3`

neuneck:
Closing because the issue is not with minio afaict now.

minio/datatypes.py
neuneck: Thank you for the pointer.

PR #1143, opened 2021-08-30
Update get_object.py

nibbl:


balamurugana (changes requested):
Send any fix to `master` branch.

PR #1152, opened 2021-09-22
Remove nose

s-t-e-v-e-n-k:
Nose has been unmaintained upstream for quite some time, sofirst, switch from nose eq_ methods to the lovely unittest assert methods. Secondly, remove nosetests from the Makefile and github actions.


balamurugana (changes requested):
Disable `C0209` at https://github.com/minio/minio-py/blob/master/Makefile#L10 and wherever applicable in this PR. We will address `C0209` as separate PR.

PR #1157, opened 2021-10-05
do not consult with certifi if use user-provided http client

pg83:


minio/api.py
```diff
@@ -139,12 +139,11 @@ def __init__(self, endpoint, access_key=None,
 
         # Load CA certificates from SSL_CERT_FILE file if set
         timeout = timedelta(minutes=5).seconds
-        ca_certs = os.environ.get('SSL_CERT_FILE') or certifi.where()
```
balamurugana, line 142: What issue are you facing here?

balamurugana (changes requested):

minio/api.py
pg83: I have custom http_client, and do not have installed certificates in my setup. So completely valid setup(I trust my private s3) just throws exceptions.

My patch delay certificate load, only if we really create urllib3.PoolManager

balamurugana: Can you share what exceptions are you getting?

pg83: Why? It can be in any range from FileNotFoundError, to (in some cases) error from Python's importlib internals, cause I use frozen python modules for distribution.

balamurugana: Makes sense. This info is needed for someone tracking PRs/issues in future.

PR #1159, opened 2021-10-19
To make possible for endpoint to contain own schema such as http:// or https://

ocworld:
This pr's purpose is making possible for endpoint to contain own schema such as http:// or https://.

First, My service using minio-py need to run in various environments that have http or https minio address in each. So, I want to handle minio address just endpoint url string without considering secure parameter.

In addition, when first time to using, considering both endpoint and secure params is confused for me. 

Therefore, I propose that supporting endpoint started with "http://" or "https". And in this case, secure is not affected.


balamurugana:
As secure is enabled by default, It is contradicting. You could wrap this behavior outside of Minio class.

PR #1165, opened 2021-11-24
fix: canonical request path if empty choose '/'

harshavardhana:
additionally this PR allows content-type and
content-length to be part of signed headers as
this is allowed in signature v4.


ravindk89 (approved):
Cursory glance looks fine.

kannappanr (approved):
LGTM

PR #1166, opened 2021-11-24
fix: do not enforce role value for replicationConfig

### LINKED ISSUE #1164: Replication Config Get issue
I'm having an issue with the bucket [get_replication_config](https://docs.min.io/docs/python-client-api-reference.html#get_bucket_replication) function. 

Environment:

I have a local [docker-compose](https://docs.min.io/docs/deploy-minio-on-docker-compose.html) which I want to replicate on a similarly deployed remote cluster. 

Issue: 

From the UI, I can without issue set the replication from one cluster to another. It works great. 

Now I wanted to code it, so first step I did was to try fetching the current replication to see what's inside a Replication Config (And frankly from an editorial side, doc is VERY lacking. I only seem to find info about the bucket class, everything else has to be learn by ourselves). 

Here's the error I get: 

image

Seems like the code need a role? But there's no where mention of a 'role' anywhere? Even in the [example](https://docs.min.io/docs/minio-bucket-replication-guide.html), the 'role' key is empty. 

I would understand for a 'set', but for a get? 

Anyway, after investigation, it seem that the issue here is in the function 'toxml' of the 'ReplicationConfig' class. 

image

Fetching the 'role' from the http reply, it doesn't seem to find any, so it tries to init it's class with an empty string. But then, raise an error at the 'if not role' line. 

---------------------------------------------------------------------------------------------------------------------------------------------------

Question: I'm I missing where the 'role' must be set? 

Else, I think there's an error in the code logic.

harshavardhana:
fixes #1164

### DIFF
--- a/minio/replicationconfig.py
+++ b/minio/replicationconfig.py
@@ -459,8 +459,6 @@ class ReplicationConfig:
     """Replication configuration."""
 
     def __init__(self, role, rules):
-        if not role:
-            raise ValueError("role must be provided")
         if not rules:
             raise ValueError("rules must be provided")
         if len(rules) > 1000:


kannappanr (approved):

minio/replicationconfig.py
```diff
@@ -459,8 +459,6 @@ class ReplicationConfig:
     """Replication configuration."""
 
     def __init__(self, role, rules):
-        if not role:
```
balamurugana, line 462: As the spec does not say this is optional, is it minio server specific behavior?

harshavardhana: The role is an AWS concept @balamurugana not applicable to us. This is our SDK as well. 

In-fact AWS SDK python works fine with MinIO. So yes it is not mandatory.

balamurugana (approved):

PR #1174, opened 2022-01-31
fix EnvAWSProvider/EnvMinioProvider to fetch access/secret keys and session token always from os.environ

nfioraio-ec:
Some providers, eg, EnvAWSProvider and EnvMinioProvider, creates the Credential object during __init__. However, if a key is None, a ValueError is raised. Therefore, when these providers are used within a ChainedProvider, the exception is raised before actually retrieving the credentials.

This bug is highlighted by the new test ChainedProviderTest.test_chain_retrieve_second, which fails without the proposed fix.

### DIFF
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -195,38 +195,29 @@ class ChainedProvider(Provider):
 class EnvAWSProvider(Provider):
     """Credential provider from AWS environment variables."""
 
-    def __init__(self):
-        self._access_key = (
-            os.environ.get("AWS_ACCESS_KEY_ID") or
-            os.environ.get("AWS_ACCESS_KEY")
-        )
-        self._secret_key = (
-            os.environ.get("AWS_SECRET_ACCESS_KEY") or
-            os.environ.get("AWS_SECRET_KEY")
-        )
-        self._session_token = os.environ.get("AWS_SESSION_TOKEN")
-
     def retrieve(self):
         """Retrieve credentials."""
         return Credentials(
-            access_key=self._access_key,
-            secret_key=self._secret_key,
-            session_token=self._session_token,
+            access_key=(
+                os.environ.get("AWS_ACCESS_KEY_ID") or
+                os.environ.get("AWS_ACCESS_KEY")
+            ),
+            secret_key=(
+                os.environ.get("AWS_SECRET_ACCESS_KEY") or
+                os.environ.get("AWS_SECRET_KEY")
+            ),
+            session_token=os.environ.get("AWS_SESSION_TOKEN"),
         )
 
 
 class EnvMinioProvider(Provider):
     """Credential provider from MinIO environment variables."""
 
-    def __init__(self):
-        self._access_key = os.environ.get("MINIO_ACCESS_KEY")
-        self._secret_key = os.environ.get("MINIO_SECRET_KEY")
-
     def retrieve(self):
         """Retrieve credentials."""
         return Credentials(
-            access_key=self._access_key,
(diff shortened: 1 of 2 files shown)


minio/credentials/providers.py
```diff
@@ -196,37 +196,38 @@ class EnvAWSProvider(Provider):
...
+        self._session_token = os.environ.get("AWS_SESSION_TOKEN")
 
     def retrieve(self):
         """Retrieve credentials."""
-        return self._credentials
+        return Credentials(
+            access_key=self._access_key,
+            secret_key=self._secret_key,
+            session_token=self._session_token,
+        )
```
balamurugana, line 215: As `os.environ` is modifiable at runtime, just fetch access/secret keys and session token every time. There is no need to store them in `__init__()`

minio/credentials/providers.py
```diff
@@ -196,37 +196,38 @@ class EnvAWSProvider(Provider):
...
-        )
+        self._access_key = os.environ.get("MINIO_ACCESS_KEY")
+        self._secret_key = os.environ.get("MINIO_SECRET_KEY")
 
     def retrieve(self):
         """Retrieve credentials."""
-        return self._credentials
+        return Credentials(
+            access_key=self._access_key,
+            secret_key=self._secret_key,
```
balamurugana, line 229: Same as above

minio/credentials/providers.py
```diff
@@ -473,11 +474,17 @@ class StaticProvider(Provider):
     """Fixed credential provider."""
 
     def __init__(self, access_key, secret_key, session_token=None):
-        self._credentials = Credentials(access_key, secret_key, session_token)
+        self._access_key = access_key
+        self._secret_key = secret_key
+        self._session_token = session_token
```
balamurugana, line 479: StaticProvider` should raise error right away for invalid access/secret keys.  It is meaningless accept invalid values here.

tests/unit/credentials_test.py
```diff
@@ -101,6 +101,26 @@ def test_chain_retrieve(self):
         self.assertEqual(creds.secret_key, "secret_aws")
         self.assertEqual(creds.session_token, "token_aws")
 
+    def test_chain_retrieve_second(self):
```
balamurugana, line 104: Please name the method appropriately.

balamurugana (changes requested):

tests/unit/credentials_test.py
nfioraio-ec: It is indeed appropriate, since we set two providers and we expect to get the second one.

minio/credentials/providers.py
nfioraio-ec: I just kept the same logic, that is, look for env vars at object creation and don't change them anymore. If we now fetch, and possibly change, them every time we may break something already running... ? Just thinking.

minio/credentials/providers.py
nfioraio-ec: Well, I don't see why it should behave differently from any other provider. Let's say we want to use EnvAWSProvider first, then AWSConfigProvider and only as a fallback the keys provided through the sw api, if any. I wouldn't prevent such application just to save a few lines of code.

balamurugana: `StaticProvider` is a commonly used credential provider. This provider is expected to return credentials without any error. This behavior cannot be changed i.e. we shouldn't return exception on `retrieve()`.

balamurugana (approved):

harshavardhana (approved):

PR #1175, opened 2022-01-31
fix finding user's home directory in AWSConfigProvider/MinioClientConfigProvider 

nfioraio-ec:
AWSConfigProvider and MinioClientConfigProvider look for config files under home directory, which is retrieved as `os.environ.get("HOME")`. However, under Windows this env var does not exists and the `os.path.join` raises a TypeError. Fixed by getting `os.environ.get("UserProfile")` or `str(Path.home())` as a last resort.

### DIFF
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -66,6 +66,7 @@ def _urlopen(http_client, method, url, body=None, headers=None):
 
 
 def _user_home_dir():
+    """Return current user home folder."""
     return (
         os.environ.get("HOME") or
         os.environ.get("UserProfile") or
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -27,6 +27,7 @@ from abc import ABCMeta, abstractmethod
 from datetime import timedelta
 from urllib.parse import urlencode, urlsplit
 from xml.etree import ElementTree
+from pathlib import Path
 
 import urllib3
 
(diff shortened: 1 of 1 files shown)


minio/credentials/providers.py
```diff
@@ -283,10 +291,14 @@ class MinioClientConfigProvider(Provider):
...
     def __init__(self, filename=None, alias=None):
+        from pathlib import Path
+
         self._filename = (
             filename or
             os.path.join(
-                os.environ.get("HOME"),
+                os.environ.get("HOME") or
+                os.environ.get("UserProfile") or
+                str(Path.home()),
```
balamurugana, line 301: Have a function to return home directory instead of repeating the logic.

minio/credentials/providers.py
```diff
@@ -283,10 +291,14 @@ class MinioClientConfigProvider(Provider):
     """Credential provider from MinIO Client configuration file."""
 
     def __init__(self, filename=None, alias=None):
+        from pathlib import Path
```
balamurugana, line 294: Please fix github CI error.

balamurugana (changes requested):

balamurugana (commented):
Please fix CI error

balamurugana (approved):

harshavardhana (approved):

PR #1182, opened 2022-02-28
Add AssumeRoleWithCertificate credential provider.

balamurugana:
Signed-off-by: Bala.FA <bala@minio.io>

### DIFF
--- a/examples/minio_with_certificate_identity_provider.py
+++ b/examples/minio_with_certificate_identity_provider.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# MinIO Python Library for Amazon S3 Compatible Cloud Storage,
+# (C) 2022 MinIO, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


harshavardhana:
@balamurugana looks like you didn't test this with MinIO. You need to provide provisions for client.crt and client.key 

You haven't added any API that takes client certs. 

minio-go API takes tls.Certificate which is a private and public key certificate - https://pkg.go.dev/crypto/tls#Certificate

harshavardhana (approved):

examples/minio_with_certificate_identity_provider.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# MinIO Python Library for Amazon S3 Compatible Cloud Storage,
+# (C) 2020 MinIO, Inc.
```
harshavardhana, line 3: should be 2022? @balamurugana

harshavardhana (changes requested):

balamurugana:
Tested the PR using self-signed certificates with ignoring certificate check using custom HTTP client.

harshavardhana (approved):

PR #1188, opened 2022-03-24
use unittest.mock instead of mock

pgajdos:
mock is not required to run the testsuite, so use built in unittest.mock.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ default: tests
 
 getdeps:
 	@echo "Installing required dependencies"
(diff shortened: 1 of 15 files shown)


pgajdos:
We can eventually make unittest.mock as a fallback, if you want.

balamurugana:
Please remove `mock` from `Makefile` and `setup.py`

balamurugana (approved):

harshavardhana (approved):

PR #1220, opened 2022-06-29
Add support for user server

taran-p:
Added main() function to some examples

Allow testing some examples on given server using environment variables

### DIFF
--- a/examples/bucket_exists.py
+++ b/examples/bucket_exists.py
@@ -14,15 +14,49 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 from minio import Minio
 
-client = Minio(
-    "play.min.io",
-    access_key="Q3AM3UQ867SPQQA43P2F",
-    secret_key="zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG",
-)
+def client_from_env()->Minio:
+    url = os.environ.get("MINIO_ADDRESS")
+    user = os.environ.get("MINIO_ACCESS_KEY")
+    pw = os.environ.get("MINIO_SECRET_KEY")
+    sec_var = os.environ.get("MINIO_SECURE",'off')
+    if sec_var == 'on':
+        sec = True
+    else:
+        sec = False
 
-if client.bucket_exists("my-bucket"):
-    print("my-bucket exists")
-else:
-    print("my-bucket does not exist")
+    if url or user or pw:
+        client = Minio(
+            url,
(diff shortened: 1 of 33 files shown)


balamurugana (changes requested):
If you want to test your server, use [functional tests](https://github.com/minio/minio-py/tree/master/tests/functional) than complicating examples.

donatello:
> If you want to test your server, use [functional tests](https://github.com/minio/minio-py/tree/master/tests/functional) than complicating examples.

@balamurugana I'm helping Taran - we're just trying to get to a situation where examples can run in CI so that they are automatically up to date with library code.

harshavardhana:
> > If you want to test your server, use [functional tests](https://github.com/minio/minio-py/tree/master/tests/functional) than complicating examples.
> 
> @balamurugana I'm helping Taran - we're just trying to get to a situation where examples can run in CI so that they are automatically up to date with library code.

If we want to do that perhaps then `examples` must have some provision to load a separate `config.py` file instead of each file being modified in the manner it's being done right now.

It kind of fails to be an example at his point if it's modified in this manner. 

My suggestion is something like this 

```
if ci: 
  from config import *
   ... use the imported values
```

donatello:
> If we want to do that perhaps then `examples` must have some provision to load a separate `config.py` file instead of each file being modified in the manner it's being done right now.
> 
> It kind of fails to be an example at his point if it's modified in this manner.
> 
> My suggestion is something like this
> 
> ```
> if ci: 
>   from config import *
>    ... use the imported values
> ```

Yeah, we're going to something like this to reduce the code in the examples - this initial change is just for some discussion and feedback. No need to merge right away.

In addition to CI, it would be good if the examples can be run locally with a user specified server rather than play - so the idea is to get this info from env vars.

balamurugana:
> > If you want to test your server, use [functional tests](https://github.com/minio/minio-py/tree/master/tests/functional) than complicating examples.
> 
> @balamurugana I'm helping Taran - we're just trying to get to a situation where examples can run in CI so that they are automatically up to date with library code.

@donatello You cannot run examples to validate the server or `minio-py` because we need to much more logic for that e.g. to test `get_bucket_versioning()`, you need an existing bucket and `set_bucket_versioning()` called in prior. We can only check these examples are valid at par with current `minio-py` APIs. This is done by `pyflakes` like tools like compilers do rather complicating examples. We should add more tests/checks in functional tests which is really helpful than these changes in examples.

PR #1225, opened 2022-07-21
select: yield available payload fully by requested num_bytes

kldtz:
When the specified number of bytes was less than the payload size, the rest of the payload was discarded. I fixed the method and adjusted the test to trigger this behavior.

### DIFF
--- a/minio/select.py
+++ b/minio/select.py
@@ -421,7 +421,7 @@ class SelectObjectReader:
         should call self.close() to release network resources.
         """
         while True:
-            if self._payload:
+            while self._payload:
                 result = self._payload
                 if num_bytes < len(self._payload):
                     result = self._payload[:num_bytes]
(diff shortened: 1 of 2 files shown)


minio/select.py
```diff
@@ -420,13 +420,10 @@ def stream(self, num_bytes=32*1024):
         Stream extracted payload from response data. Upon completion, caller
         should call self.close() to release network resources.
         """
-        while True:
-            if self._payload:
+        while self._read() > 0:
```
balamurugana, line 423: This problem never occurs because of https://github.com/minio/minio-py/blob/master/minio/select.py#L273. You will get an exception. If `RawIOBase.read()` returns less bytes than requested, but still not received EOF, you would need to fix `_read()` function.

balamurugana (changes requested):

minio/select.py
kldtz: The `_read()` function works correctly, as far as I can tell. The bug is in the `stream` method. If you run the adjusted test (with `num_bytes=16`, which is less than the payload size), you'll see that it fails without the fix.

balamurugana (approved):

harshavardhana (approved):

PR #1240, opened 2022-09-26
list_objects: fix parsing user metadata as per MinIO server

harshavardhana:


minio/datatypes.py
```diff
@@ -214,11 +214,14 @@ def fromxml(cls, element, bucket_name, is_delete_marker=False,
...
-        for child in tag:
-            key = child.tag.split("}")[1] if "}" in child.tag else child.tag
-            metadata[key] = child.text
+        selement = find(element, "UserMetadata")
+        if selement:
+            for child in findall(selement, "Items"):
+                key = find(child, "Key")
+                value = find(child, "Value")
+                if key and value:
+                    metadata[key.text] = value.text
```
balamurugana, line 224: You could directly use `metadata[findtext(child, "Key", True)] =  findtext(child, "Value", True)`

harshavardhana: let me do that @balamurugana

harshavardhana:
PTAL @balamurugana

balamurugana (approved):

renaudcalmont:
Not sure what happened here but the code is clearly trying to address incompatible cases at once... the commit that was reverted was solving one of these, it shoud be re-implemented somehow

PR #1252, opened 2022-12-07
Bucket history feature

archx-lab:
I'm trying to add the functionality for monitoring a static website and monitoring S3 Lifecycle activity. Two types of log files can be created: a Log file which records logs of all the buckets associated with the particular client (super-log-file) and a Log file which records logs of all the objects associated with the specific bucket of the particular client (sub-log-file).


balamurugana:
Do not send your custom requirements.

Issue #1261, opened 2023-01-11
Handling "Bucket name does not follow S3 standards" for trailing dash


cal-pratt:
I've been using the boto libraries for a while and wanted to switch over to using the Minio library for the simpler interface.
We're using a self hosted solution implementing S3 (using Ceph, not Minio server or AWS S3).

One thing we've caught is that some of our bucket names are being rejected because they end with a dash `-`, by `check_bucket_name` https://github.com/minio/minio-py/blob/039ff29d76a45427498d04153ddc74776f5d777f/minio/helpers.py#L203-L213

I understand that ending with a dash is allowed in the proper S3 spec, but this is how our buckets are currently configured and it would be a significant amount of engineering work to change it. I was wondering if we could allow ending with a dash for the non-strict regex? That would help a lot!

-------

PS. I've tested with the following hack and everything seems to work how I would expect:

```py
from minio import Minio
import minio.helpers
minio.helpers._VALID_BUCKETNAME_REGEX = re.compile(
    '^[A-Za-z0-9][A-Za-z0-9\\.\\-\\_\\:]{1,61}[A-Za-z0-9-]$'  # allowed to end with -
)

client = Minio(
    endpoint=my_endpoint,
    access_key=my_access_key,
    secret_key=my_secret_key,
)
client.fget_object("some-bucket-", "some.file", "some.file")
```

cal-pratt:
Another option that might be less invasive to other users is to allow users to provide their own validation regex's as part of the `Minio.__init__` method.

balamurugana:
[Bucket naming rules](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html) does not allow `-` at the end and we don't want to support non-standard names.

You could inherit `Minio` class for all custom requirements.

cal-pratt:
@balamurugana this seems to be hardcoded in the module constant. Could you explain how I could avoid it with inheritance?

PR #1262, opened 2023-01-16
Allow customizable check_bucket_name behavior

cal-pratt:
Following up on https://github.com/minio/minio-py/issues/1261 the suggestion given was to inherit the Minio class for all custom requirements. In this case I want to override the behavior of `check_bucket_name` to support my legacy usecase. Because this is a module level helper function, I couldn't easily replace it with inheritance, so this change is moving references to the helper into a proxy method on the instance.

This change allows me to override the default bucket checking logic easily without having to loosen the strict checks for other clients:

```py
BUCKET_NAME_EXEMPTIONS = ["legacy-bucket-ending-in-dash-"]

class CustomMinio(Minio):
    def _check_bucket_name(self, bucket_name, strict=False):
        if bucket_name not in BUCKET_NAME_EXEMPTIONS:
            return super()._check_bucket_name(bucket_name, strict)
```

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -561,7 +561,7 @@ class Minio:  # pylint: disable=too-many-public-methods
                     print(data.decode())
                 print(result.stats())
         """
-        check_bucket_name(bucket_name)
+        self._check_bucket_name(bucket_name)
         check_non_empty_string(object_name)
         if not isinstance(request, SelectRequest):
             raise ValueError("request must be SelectRequest type")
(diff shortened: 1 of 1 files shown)


balamurugana:
As I mentioned, we do not support non-standard behavior. You could use monkeypatching or dealing with your own fork.

cal-pratt:
> As I mentioned, we do not support non-standard behavior. You could use monkeypatching or dealing with your own fork.

@balamurugana Could you give the code I wrote another look? It really is a non-invasive change. Monkeypatching in production code sounds awful, and maintaining a fork is a non-trivial amount of ongoing future work. This PR is really just a refactoring that allows me to get away without anything gross like those alternatives. Otherwise I'll unfortunately have to take a pass on using this library, which seems like a shame.

cal-pratt:
Additionally, with this change I could also see it being used support being more restrictive.
e.g. for a shared s3 instance to support having a common prefix to all of my bucket names. Would this use-case be okay if updated with supporting documentation for that purpose?

PR #1272, opened 2023-03-30
fix policy as a string in presigned_post_policy() API

cniackz:
### Objective:

Provide working example for `examples/presigned_post_policy.py` as current one isn't working due to improper filled form being provided in the curl that we print as example:

```
-F policy=b'eyJleHBpcmF0aW9uIjogIjIwMjMtMDQtMDhUMTY6MDY6MzkuOTQyWiIsICJjb25kaXRpb25zIjogW1siZXEiLCAiJGJ1Y2tldCIsICJteS1idWNrZXQiXSwgWyJzdGFydHMtd2l0aCIsICIka2V5IiwgIm15L29iamVjdC9wcmVmaXgvIl0sIFsiY29udGVudC1sZW5ndGgtcmFuZ2UiLCAxMDQ4NTc2LCAxMDQ4NTc2MF0sIFsiZXEiLCAiJHgtYW16LWFsZ29yaXRobSIsICJBV1M0LUhNQUMtU0hBMjU2Il0sIFsiZXEiLCAiJHgtYW16LWNyZWRlbnRpYWwiLCAiUTNBTTNVUTg2N1NQUVFBNDNQMkYvMjAyMzAzMjkvdXMtZWFzdC0xL3MzL2F3czRfcmVxdWVzdCJdLCBbImVxIiwgIiR4LWFtei1kYXRlIiwgIjIwMjMwMzI5VDE2MDY0MFoiXV19'
```

> Causing

```
Code: SignatureDoesNotMatch
Message: The request signature we calculated does not match the signature you provided. Check your key and signing method.
```

What I suggest instead is a fully working example for user to get this working fast and easy

### Tested:

Screenshot 2023-03-29 at 9 02 48 PM


examples/presigned_post_policy.py
```diff
@@ -25,19 +26,36 @@
...
     "https://play.min.io/my-bucket "
-    "{0} -F file=@<FILE>"
+    "{0} -F file=@output.txt -F key=my/object/prefix/my-object-name -F Content-Type=text/html -v"
 ).format(
     " ".join(["-F {0}={1}".format(k, v) for k, v in form_data.items()]),
 )
 print(curl_cmd)
+
+# Execute the curl for the user:
+os.system(curl_cmd)
```
balamurugana, line 61: The idea of this example is how to use `presigned_post_policy()` API. If executing a POST call wih form data is needed, you could use `urllib3` than external `curl` binary.

examples/presigned_post_policy.py
```diff
@@ -25,19 +26,36 @@
...
 )
 
+# In play make sure my-bucket exist if not please create it
 policy = PostPolicy(
     "my-bucket", datetime.utcnow() + timedelta(days=10),
 )
-policy.add_starts_with_condition("key", "my/object/prefix/")
+# key=my/object/prefix/my-object-name should go in the policy and in the filled-in form
+# Same as in our Java example for the example to actually works.
+policy.add_starts_with_condition("key", "my/object/prefix/my-object-name")
```
balamurugana, line 35: `add_starts_with_condition` should work with `my/object/prefix/`. If not, it is a server bug.

examples/presigned_post_policy.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # MinIO Python Library for Amazon S3 Compatible Cloud Storage,
-# (C) 2015 MinIO, Inc.
+# (C) 2023 MinIO, Inc.
```
balamurugana, line 3: License could be `2015` to `2023`

cniackz: Ok @balamurugana it make sense, I have changed this accordingly

examples/presigned_post_policy.py
```diff
@@ -25,19 +26,36 @@
     secret_key="zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG",
 )
 
+# In play make sure my-bucket exist if not please create it
```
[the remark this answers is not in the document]
cniackz: Removed

examples/presigned_post_policy.py
```diff
@@ -33,10 +33,21 @@
 
 form_data = client.presigned_post_policy(policy)
 
+# Decode from bytes so POST will work by having the policy as policy=eyJleHBpcm
+# If policy is like "policy=b'eyJleHBpcm..." Then this will fail.
+if isinstance(form_data['policy'], (bytes, bytearray)):
+    form_data['policy'] = form_data['policy'].decode("utf-8")
```
balamurugana, line 39: value of `form_data["policy"]` should contain string. This bug needs to fixed at https://github.com/minio/minio-py/blob/master/minio/datatypes.py#L777 like `policy = base64.b64encode(json.dumps(policy).encode()).decode("utf-8")`

examples/presigned_post_policy.py
```diff
@@ -33,10 +33,21 @@
...
+
+# Create the file for the example
+# 2 MB Object to be in the defined range above in the policy.
+with open("output.txt", "wb") as out:
+    out.truncate(1 * 1024 * 1024)
```
balamurugana, line 44: As we are not executing `curl` command, this fie creation is not required.

examples/presigned_post_policy.py
cniackz: Oh wow, you are right, this is much better than just converting in the example!!!.

Thanks Bala, really appreciate it, let fix the bug in a separate PR then.

examples/presigned_post_policy.py
```diff
@@ -33,10 +33,21 @@
...
+
+# This example is made with text file, please change file and content type accordingly for an image.
 curl_cmd = (
     "curl -X POST "
     "https://play.min.io/my-bucket "
-    "{0} -F file=@<FILE>"
+    "{0} -F file=@output.txt -F key=my/object/prefix/ -F Content-Type=text/html -v"
```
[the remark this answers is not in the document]
cniackz: I see, I believe you are right, let me just test it locally one time.

examples/presigned_post_policy.py
cniackz: Yeah it can be removed to keep example generic, will do

examples/presigned_post_policy.py
cniackz: Let's fix it in this PR.

examples/presigned_post_policy.py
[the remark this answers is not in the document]
cniackz: Nope, this is not going to work Bala:
```
<Error><Code>MethodNotAllowed</Code><Message>The specified method is not allowed against this resource.</Message><BucketName>my-bucket</BucketName><Resource>/my-bucket/my/object/prefix</Resource><Region>us-east-1</Region><RequestId>17513C2C0783964D</RequestId><HostId>e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855</HostId></Error>Cesars-MacBook-Pro:examples cniackz$ 
```

It has to be the bucket

examples/presigned_post_policy.py
cniackz: Removed.

examples/presigned_post_policy.py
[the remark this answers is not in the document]
balamurugana: Actually we need to pass `-F key=<OBJECT-NAME>` and the request is to the bucket.

minio/datatypes.py
```diff
@@ -775,8 +775,12 @@ def form_data(self, creds, region):
         policy["conditions"].append([_EQ, "$x-amz-date", amz_date])
 
         policy = base64.b64encode(json.dumps(policy).encode())
+        if isinstance(policy, (bytes, bytearray)):
+            policy = policy.decode()  # convert policy to string
+        if isinstance(policy, str) is not True:
```
balamurugana, line 780: As `base64.b64encode()` always return `bytes`, you would need to directly decode to string. It is better to fail than type checking in python.

cniackz: ok, it make sense, thank you Bala

balamurugana (approved):

harshavardhana (approved):

captainIT:
1、run  presigned_post_policy.py
2、get result
`curl -X POST https://play.min.io/my-bucket -F x-amz-algorithm=AWS4-HMAC-SHA256 -F x-amz-credential=Q3AM3UQ867SPQQA43P2F/20230524/us-east-1/s3/aws4_request -F x-amz-date=20230524T095400Z -F policy=b'eyJleHBpcmF0aW9uIjogIjIwMjMtMDYtMDNUMDk6NTM6NDkuOTM3WiIsICJjb25kaXRpb25zIjogW1siZXEiLCAiJGJ1Y2tldCIsICJteS1idWNrZXQiXSwgWyJzdGFydHMtd2l0aCIsICIka2V5IiwgIm15L29iamVjdC9wcmVmaXgvIl0sIFsiY29udGVudC1sZW5ndGgtcmFuZ2UiLCAxMDQ4NTc2LCAxMDQ4NTc2MF0sIFsiZXEiLCAiJHgtYW16LWFsZ29yaXRobSIsICJBV1M0LUhNQUMtU0hBMjU2Il0sIFsiZXEiLCAiJHgtYW16LWNyZWRlbnRpYWwiLCAiUTNBTTNVUTg2N1NQUVFBNDNQMkYvMjAyMzA1MjQvdXMtZWFzdC0xL3MzL2F3czRfcmVxdWVzdCJdLCBbImVxIiwgIiR4LWFtei1kYXRlIiwgIjIwMjMwNTI0VDA5NTQwMFoiXV19' -F x-amz-signature=860ea4a6b5417c2e708b1e0c85f5fcf6e68ac3d0db1eea57a26555d3b185f4d7 -F file=@<FILE> -F key=<OBJECT-NAME>`

3、get  error after run result
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>MalformedPOSTRequest</Code><Message>The body of your POST request is not well-formed multipart/form-data. (The name of the uploaded key is missing)</Message><BucketName>my-bucket</BucketName><Resource>/my-bucket</Resource><Region>us-east-1</Region><RequestId>17620BAFF03D8D43</RequestId><HostId>3e996b2f640d7e065d3a5c4e39a5538cefb82e3e77771990265e4698d8681eac</HostId></Error>

PR #1286, opened 2023-06-26
Fix upload snowball objects with staging file

emrocha:
There is a bug in upload_snowball_objects when it is called with "staging_filename" parameter. The variable "length" passed to fput_object is not defined in this case.

### DIFF
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -1899,7 +1899,8 @@ def test_upload_snowball_objects(log_entry):
     _test_upload_snowball_objects(log_entry)
 
 
-def test_upload_snowball_objects_with_staging(log_entry):
+def test_upload_snowball_objects_with_staging( # pylint: disable=invalid-name
+        log_entry):
     """Test upload_snowball_objects() with staging file."""
     staging_filename = f"{uuid4()}.tar"
     _test_upload_snowball_objects(log_entry, staging_filename)
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -1856,7 +1856,7 @@ def test_remove_bucket(log_entry):
     _CLIENT.remove_bucket(bucket_name)
 
 
-def test_upload_snowball_objects(log_entry):
+def _test_upload_snowball_objects(log_entry, staging_filename=None):
     """Test upload_snowball_objects()."""
 
     # Get a unique bucket_name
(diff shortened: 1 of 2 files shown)


tests/functional/tests.py
```diff
@@ -1891,6 +1891,32 @@ def test_upload_snowball_objects(log_entry):
         _CLIENT.remove_object(bucket_name, "my-object3")
         _CLIENT.remove_bucket(bucket_name)
 
+    # run with staging file option
```
balamurugana, line 1894: Have `def _test_upload_snowball_objects(log_entry, staging_filename):` and use the method to have two tests named `test_upload_snowball_objects()` and `test_upload_snowball_objects_staging_file` to avoid code duplication.

tests/functional/tests.py
```diff
@@ -1891,6 +1891,32 @@ def test_upload_snowball_objects(log_entry):
...
+                SnowballObject("my-object1", data=io.BytesIO(b"py"), length=2),
+                SnowballObject(
+                    "my-object2", data=reader1, length=size,
+                ),
+                SnowballObject(
+                    "my-object3", data=reader2, length=size,
+                    mod_time=datetime.now(),
+                ),
+            ],
+            staging_filename="staging.tar"
```
balamurugana, line 1912: Use random filename here like ` f"{uuid4()}.tar"`

balamurugana (changes requested):

tests/functional/tests.py
```diff
@@ -1883,13 +1883,26 @@ def test_upload_snowball_objects(log_entry):
...
         _CLIENT.remove_bucket(bucket_name)
+        if staging_filename and os.path.exists(staging_filename):
+            os.remove(staging_filename)
+
+def test_upload_snowball_objects(log_entry):
+    """Test upload_snowball_objects()."""
+    _test_upload_snowball_objects(log_entry)
+
+
+def test_upload_snowball_objects_with_staging(log_entry):
```
balamurugana, line 1902: Add in-line comment `  # pylint: disable=invalid-name`  to ignore this pylint error.

balamurugana (changes requested):

balamurugana (approved):

harshavardhana (approved):

Issue #1243, opened 2022-10-26
MinioAdmin needs mc installed on local machine


Rocking80:
I checked `MinioAdmin` class in `minioadmin` file. It seems for admin we can only execute admin commands instead of the rest api way as `Minio`? So this needs `mc` executable to be installed and configured on local machine, right? This is not convenient and has much limitation in automation.

balamurugana:
Yep. It requires `mc` on local machine. Please feel free to send a PR to use `urllib3`.

Issue #829, opened 2019-12-07
Python admin api


ruiztulio:
Hi, is it possible to access the admin api via the python package? I don't find it in the documetation neither the code (probbably not looking in the right place).

I'm trying to write an app that uses minio. The users, groups and policies will be managed from there. I know I can do it using minio client, but it'd be nice to handle everything from the api itself

If there is no way to do it from the minio-py package, the admin api is documented anywhere? I could only find the mc

Thanks for the work dedicated to this package btw!

harshavardhana:
It is not possible yet, we may do it in the future but not for now.

ruiztulio:
@harshavardhana thanks for the answer...is it documented somewhere?  I wouldn't mind to implement it by myself

harshavardhana:
https://github.com/minio/minio/tree/master/pkg/madmin @ruiztulio

ruiztulio:
@harshavardhana Thanks! I'll close this then

jhadjar:
Hey, @ruiztulio... Did you end up implementing this? We implemented something in that sense here https://big-mama-tech.gitlab.io/bmc/. Would this API suit you? No external libraries, provided you have the `mc` and `minio` command line tools on your machine.

It's still a prototype for our work for https://iko.ai and does not yet completely match `mc` and `minio` CLIs but we'll probably beef it up and share it.

@harshavardhana Is this on the roadmap?

harshavardhana:
> Hey, @ruiztulio... Did you end up implementing this? We implemented something in that sense here https://big-mama-tech.gitlab.io/bmc/. Would this API suit you? No external libraries, provided you have the `mc` and `minio` command line tools on your machine.
> 

This is correct usage.

> It's still a prototype for our work for https://iko.ai and does not yet completely match `mc` and `minio` CLIs but we'll probably beef it up and share it.
> 
> @harshavardhana Is this on the roadmap?

No @jhadjar

ThyMakra:
Hi. @jhadjar that was very helpful though is there anyway to use `bmc` to manipulate `mc admin` to configure the bucket (for example: `bucket quota`)

ruiztulio:
@jhadjar Hi, sorry for reaching you so late, but I did not implement it, actually a few months ago revisited the thread and started using bmc, just works great so far! Cheers!

jhadjar:
[...]

PR #1291, opened 2023-07-25
Refactor MinioAdminClient using HTTP client

pbrw:
Attempt to solve these issues #829 #1243

- Implemented encryption/decryption of sensitive payload which is the hard part about using the HTTP Admin API
- Refactored code to reuse the request authentication functionality 
- Implemented basic user management methods (user_add, user_remove, user_info, list_users) 
- Wrote some tests

Now it should be straightforward to implement the rest of MinIO Admin API


minio/crypto.py
```diff
@@ -0,0 +1,112 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# pylint: disable=too-many-lines,disable=too-many-branches,too-many-statements
+# pylint: disable=too-many-arguments
+
+"""Cryptography to read and write encrypted MinIO Admin payload"""
+
+import os
```
balamurugana, line 22: Can't we import `urandom` only?

minio/minioadminhttp.py
```diff
@@ -0,0 +1,108 @@
...
+from minio.api import HttpClient
+from minio.crypto import decrypt, encrypt
+from minio.datatypes import UserInfo, parse_list_users
+from minio.error import AdminResponseError
+
+
+_ADMIN_PATH_PREFIX = "/minio/admin/v3"
+
+
+class MinioAdminHttp(HttpClient):
```
balamurugana, line 33: Move code in this class to `MinioAdmin` class and remove `mc` usage in the class.

minio/helpers.py
```diff
@@ -512,9 +512,9 @@ def virtual_style_flag(self, flag):
         """Check to use virtual style or not."""
         self._virtual_style_flag = flag
 
-    def build(
-            self, method, region,
-            bucket_name=None, object_name=None, query_params=None,
+    def build(  # pylint: disable=too-many-branches
```
balamurugana, line 515: `BaseUrl` is meant only for Minio S3 client. For AdminClient, we just need

```py
class MinioAdmin:
    def __init__(self, host, credentials, port=9000, secure=True, region="', http_client=None, cert_check=True):
```

minio/error.py
```diff
@@ -53,6 +53,22 @@ def __reduce__(self):
...
 
+class AdminResponseError(MinioException):
+    """Raised to indicate that non-OK response from server."""
+
+    def __init__(self, code, content_type, body):
+        self._code = code
+        self._content_type = content_type
+        self._body = body
+        super().__init__(
+            f"non-OK admin response from server; "
```
balamurugana, line 64: Message should be `error response from server; `

minio/api.py
```diff
@@ -34,46 +33,40 @@
...
 from .xml import Element, SubElement, findtext, getbytes, marshal, unmarshal
 
-_DEFAULT_USER_AGENT = (
-    f"MinIO ({platform.system()}; {platform.machine()}) "
-    f"{__title__}/{__version__}"
-)
-
 
-class Minio:  # pylint: disable=too-many-public-methods
+class Minio(HttpClient):  # pylint: disable=too-many-public-methods
```
balamurugana, line 69: Nothing needs to be changed in this class. You would need to have separate methods inside `MinioAdmin` class.

minio/__init__.py
```diff
@@ -41,3 +41,4 @@
 from .api import Minio
 from .error import InvalidResponseError, S3Error, ServerError
 from .minioadmin import MinioAdmin
+from .minioadminhttp import MinioAdminHttp
```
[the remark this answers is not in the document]
pbrw: Can you elaborate? I can use minio admin as `from minio import Minio, MinioAdminHttp`

minio/minioadminhttp.py
pbrw: Should I change the interface of the MinioAdmin class and make a breaking change?
Also, there is a lot of user commands and I wanted to implement them step by step. Is it okay to make only the subset of them available at the moment?

minio/http.py
```diff
@@ -0,0 +1,167 @@
...
+from .helpers import (BaseURL, md5sum_hash, sha256_hash)
+
+from .signer import sign_v4_s3
+
+_DEFAULT_USER_AGENT = (
+    f"MinIO ({platform.system()}; {platform.machine()}) "
+    f"{__title__}/{__version__}")
+
+
+class HttpClient:
```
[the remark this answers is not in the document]
pbrw: You suggested before is to have a common base class for `Minio` and `MinioAdmin` to make HTTP requests. I can change it to reuse only the content of `helpers.py`

minio/api.py
pbrw: Ok, but you suggested to make a base class before

minio/minioadminhttp.py
balamurugana: Yes. There is no need to have a separate class. We have to leverage `MinioAdmin` class. APIs in `MinioAdmin` class return `dict`. You could keep the APIs returning `dict`.

We would need to have current APIs available by http client. May be you could send additional changes to this PR as new commits.

pbrw: Ok, I get it. I will change this PR and remove changes to api.py.

Do I need to implement all methods in the current API or can I start with only the basic ones?

balamurugana: We need to have all methods, but you could implement one after another in this PR as a new commits

pbrw:
I rewrote it a little bit according to your requests. At the moment I commented all the MinioAdmin commands with `mc` and I will rewrite them with HTTP in subsequent commits

balamurugana:
> I rewrote it according to your requests. At the moment I commented all the MinioAdmin commands with `mc` and I will rewrite them with HTTP in subsequent commits

@pbrw I sent a PR https://github.com/pbrw/minio-py/pull/1 to your fork instead of lengthy review. Please review and merge it in your fork.

balamurugana:
@pbrw I sent you PR https://github.com/pbrw/minio-py/pull/3 and reverted `prometheus_create()` addition. Few APIs are removed and site replication APIs are added. I also increased minor version to indicate backward compatibility.

minio/minioadmin.py
```diff
@@ -538,9 +539,33 @@ def top_locks(self):
...
+        default_job_name = "minio-job"
+        default_metrics_path = "/minio/v2/metrics/cluster"
+        expire_time = datetime.utcnow() + timedelta(days=default_expire_time)
+        token = jwt.encode({
+            "iss": "prometheus",
+            "sub": self._provider.retrieve().access_key,
+            "exp": expire_time
+        }, self._provider.retrieve().secret_key, algorithm="HS512")
+
+        return {
```
harshavardhana: This must be a yaml not a dictionary, but this doesn't warrant a minio-py API. Since this doesn't involve any network call at all.

harshavardhana: I would suggest we leave this out of admin API for now

harshavardhana (changes requested):

balamurugana:
@pbrw Please fix CI failure

balamurugana (approved):

harshavardhana (approved):

PR #1296, opened 2023-08-28
Remove 'config_remove' admin command

pbrw:
According to [docs](https://min.io/docs/minio/linux/reference/minio-mc-admin/mc-admin-config.html) command `mc admin config remove` no longer exists.


balamurugana:
This PR is no more needed due to https://github.com/minio/minio-py/pull/1291

PR #1302, opened 2023-08-31
add mypy

trim21:
add mypy to check type error

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ default: tests
 
 getdeps:
 	@echo "Installing required dependencies"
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -3,14 +3,15 @@ default: tests
 
 getdeps:
 	@echo "Installing required dependencies"
-	@pip install --user --upgrade autopep8 certifi pytest pylint urllib3
+	@pip install --user --upgrade autopep8 certifi pytest pylint urllib3 mypy
```
balamurugana, line 6: We don't need this linter. This is not a requirement now.

balamurugana (changes requested):

PR #1303, opened 2023-08-31
Format with black

trim21:
autopep8 doesn't enforce code style in some corner case, use black to make sure formatter of different editor/ide won't cause unnecessary diff.


balamurugana (changes requested):
No. We don't need to use `black`. You have to fix it at your side.

PR #1305, opened 2023-09-06
Add ability to use tqdm.tqdm as progress bar

balbok0:
Since tqdm is a de-facto progress bar library in python, this adds ability to display it as progress bar in `put_object`, `fput_object` and `fget_object`.

My understanding is that dependencies of minio python client should be minimal, so no formal typing/isinstance checks are provided.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -1005,7 +1005,7 @@ class Minio:  # pylint: disable=too-many-public-methods
         :param version_id: Version-ID of the object.
         :param extra_query_params: Extra query parameters for advanced usage.
         :param tmp_file_path: Path to a temporary file.
-        :param progress: A progress object
+        :param progress: A Progress or tqdm.tqdm object
         :return: Object information.
 
         Example::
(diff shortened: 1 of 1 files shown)


minio/api.py
```diff
@@ -1057,9 +1057,17 @@ def fget_object(self, bucket_name, object_name, file_path,
...
 
             if progress:
-                # Set progress bar length and object name before upload
+                # Set progress bar length and object name before download
                 length = int(response.headers.get('content-length', 0))
-                progress.set_meta(object_name=object_name, total_length=length)
+                if not isinstance(progress, Thread):
+                    # Set progress bar length and object name before upload
+                    progress.set_meta(object_name=object_name, total_length=length)
+                elif hasattr(progress, "set_description") and hasattr(progress, "reset"):
```
balamurugana, line 1065: We could have generic progress bar support instead of this.
```py
class Progress:
    """A generic progress class."""
    __metaclass__ = ABCMeta

    @abstractmethod
    def set_meta(self, object_name, total_length):
        """Set progress metadata."""

    @abstractmethod
    def update(self, length):
        """Update progress."""
```
Here, check for `progress` is instance of `Progress`.

balamurugana (changes requested):

minio/api.py
balbok0: `Progress` is only defined in examples for now.

I can define an abstract class instead, but I would follow this spec:
```
class Progress:
    """A generic progress class."""
    __metaclass__ = ABCMeta

    @abstractmethod
    def set_description(self, desc: str):
        """Set progress description."""

    @abstractmethod
    def reset(self, total: int):
        """Set progress description."""

    @abstractmethod
    def update(self, length):
        """Update progress."""
```

balamurugana: This breaks backward compatibility which we don't want to do that. `tqdm` users can wrap `set_description()` and `reset()` into `set_meta()` of their own wrapper class.

harshavardhana:
Please resubmit the PR if this is still valid.

PR #1311, opened 2023-10-05
fix remove_objects() example to convert map to list of map

dsgibbons:
`Minio.remove_objects` expects a list of `DeleteObject`s, but the documentation uses a `map`. If a `map` is passed to `Minio.remove_objects`, it will silently fail, and the objects will not be deleted. This PR corrects the documentation to pass a `list` instead of a `map` to `Minio.remove_objects`.

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -2002,10 +2002,10 @@ class Minio:  # pylint: disable=too-many-public-methods
                 print("error occurred when deleting object", error)
 
             # Remove a prefix recursively.
-            delete_object_list = map(
+            delete_object_list = list(map(
                 lambda x: DeleteObject(x.object_name),
                 client.list_objects("my-bucket", "my/prefix/", recursive=True),
-            )
+            ))
             errors = client.remove_objects("my-bucket", delete_object_list)
             for error in errors:
                 print("error occurred when deleting object", error)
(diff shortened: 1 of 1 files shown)


minio/api.py
```diff
@@ -2002,10 +2002,14 @@ def remove_objects(self, bucket_name, delete_object_list,
                 print("error occurred when deleting object", error)
 
             # Remove a prefix recursively.
-            delete_object_list = map(
-                lambda x: DeleteObject(x.object_name),
-                client.list_objects("my-bucket", "my/prefix/", recursive=True),
-            )
+            delete_object_list = [
+                DeleteObject(x).object_name
+                for x in client.list_objects(
```
balamurugana, line 2007: What problem did you get when using `map()`?

balamurugana (changes requested):

minio/api.py
dsgibbons: `list(map(...))` works fine, but just using `map` doesn't work with `Minio.remove_objects`. This method expects a list. Passing a `map` fails silently and the objects are not removed.

In the above commits, I changed from `list(map(...))` to a list comprehension to make the code a bit more concise.

balamurugana: That is too verbose. Please make `list(map())`

dsgibbons: Reverted to `list(map())`

balamurugana (approved):

PR #1320, opened 2023-10-20
Add a py.typed file to expose type annotations

brianhelba:
See: https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/


trim21:
should we add this file after package is fully typed？

brianhelba:
I think it'd be beneficial to add now, so downstream packages can start to use the annotations which already exist. Anything without an annotation will just default to `Any`, which is what happens right now (without `py.typed`).

I think enabling this might actually help to encourage community contributions to help get this package fully typed, since downstream users will want to add annotations to the APIs which they use.

balamurugana:
Refer https://github.com/minio/minio-py/pull/1376

PR #1321, opened 2023-10-20
Allow duck typing for Progress objects

brianhelba:
It's not strictly necessary for `Progress` objects to be Threads, as none of the `Thread` APIs are actually used. Rather, `Progress` must only be an object which implements `set_meta` and `update` methods.

To support (optional) type checking of `Progress`, this also adds a `Progress` protocol, which consumers may choose (but are not required) to subclass.


balamurugana:
Already addressed

PR #1323, opened 2023-10-20
Drop support for Python 3.7 and add support for 3.12

brianhelba:
[Python 3.7 is now end-of-life](https://devguide.python.org/versions/).


balamurugana:
Adding Python v3.12 support at https://github.com/minio/minio-py/pull/1338

PR #1326, opened 2023-11-03
Update the Python Quickstart README

feorlen:
Improve the Python Quickstart:

- Add a minimal client example
- Clarify the steps to run the file uploader example.
- Update text and formatting to more closely match the MinIO docs.

Since this repo doesn't use the Docs staging server, this direct link in GitHub is an easier way to see an overview of changes than reading the diff:
https://github.com/feorlen/minio-py/blob/readme-quickstart-update/README.md

This update roughly follows the structure of the `minio-go` [README](https://github.com/minio/minio-go/blob/master/README.md).

Corresponding docs issue: https://github.com/minio/docs/issues/1042

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,9 @@
+MinIO Cloud Storage, (C) 2014-2023 MinIO, Inc.
+
+This product includes software developed at MinIO, Inc.
+(https://min.io/).
+
+The MinIO project contains unmodified/modified subcomponents too with
+separate copyright notices and license terms. Your use of the source
+code for these subcomponents is subject to the terms and conditions
+of Apache License Version 2.0
--- a/README.md
+++ b/README.md
@@ -1,19 +1,21 @@
-# MinIO Python SDK for Amazon S3 Compatible Cloud Storage [![Slack](https://slack.min.io/slack?type=svg)](https://slack.min.io)
+# MinIO Python Client SDK for Amazon S3 Compatible Cloud Storage [![Slack](https://slack.min.io/slack?type=svg)](https://slack.min.io) [![Sourcegraph](https://sourcegraph.com/github.com/minio/minio-py/-/badge.svg)](https://sourcegraph.com/github.com/minio/minio-py?badge) [![Apache V2 License](https://img.shields.io/badge/license-Apache%20V2-blue.svg)](https://github.com/minio/minio-py/blob/master/LICENSE)
 
-MinIO Python SDK is Simple Storage Service (aka S3) client to perform bucket and object operations to any Amazon S3 compatible object storage service.
+The MinIO Python Client SDK provides straightforward APIs to access any Amazon S3 compatible object storage.
 
-For a complete list of APIs and examples, please take a look at the [Python Client API Reference](https://min.io/docs/minio/linux/developers/python/API.html)
+This Quickstart Guide covers how to install the MinIO client SDK, connect to MinIO, and create a sample file uploader.
+For a complete list of APIs and examples, see the [Python Client API Reference](https://min.io/docs/minio/linux/developers/python/API.html)
 
-## Minimum Requirements
-Python 3.7 or higher.
+These examples presume a working [3.7+ Python development environment](https://www.python.org/downloads/) and the [MinIO `mc` command line tool](https://min.io/docs/minio/linux/reference/minio-mc.html).
 
-## Download using pip
+## Install the Minio Python SDK
+
+### Using pip
 
 ```sh
 pip3 install minio
 ```
 
-## Download source
+### From GitHub
 
 ```sh
 git clone https://github.com/minio/minio-py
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -1,69 +1,115 @@
-# MinIO Python SDK for Amazon S3 Compatible Cloud Storage [![Slack](https://slack.min.io/slack?type=svg)](https://slack.min.io)
+# MinIO Python Client SDK for Amazon S3 Compatible Cloud Storage [![Slack](https://slack.min.io/slack?type=svg)](https://slack.min.io) [![Sourcegraph](https://sourcegraph.com/github.com/minio/minio-py/-/badge.svg)](https://sourcegraph.com/github.com/minio/minio-py?badge) [![Apache V2 License](https://img.shields.io/badge/license-Apache%20V2-blue.svg)](https://github.com/minio/minio-py/blob/master/LICENSE)
```
feorlen, line 1: The minio-go README includes these, added here as well. Maybe the "used by" (Sourcegraph) is incorrect? Says 0 projects.

README.md
```diff
@@ -1,69 +1,115 @@
...
+For a complete list of APIs and examples, see the [Python Client API Reference](https://min.io/docs/minio/linux/developers/python/API.html)
 
-## Minimum Requirements
-Python 3.7 or higher.
+These examples presume a working [3.7+ Python development environment](https://www.python.org/downloads/) and the [MinIO `mc` command line tool](https://min.io/docs/minio/linux/reference/minio-mc.html).
 
-## Download using pip
+## Install the Minio Python SDK
+
+### Using pip
```
feorlen, line 12: The original appeared to suggest you had to install both with pip and from source. This didn't seem necessary when I tried, so I made them sibling headings. Is this accurate?

README.md
```diff
@@ -1,69 +1,115 @@
...
 git clone https://github.com/minio/minio-py
 cd minio-py
 python setup.py install
 ```
 
-## Quick Start Example - File Uploader
-This example program connects to an S3-compatible object storage server, make a bucket on that server, and upload a file to the bucket.
-
-You need the following items to connect to an S3-compatible object storage server:
+## Initialize a MinIO Client Object
```
feorlen, line 26: Added a similar minimal example, as the Go one does.

README.md
```diff
@@ -1,69 +1,115 @@
...
+def main():
+    # Create a client with the MinIO server playground, its access key          
+    # and secret key.                                                           
+    client = Minio(
+        endpoint="play.min.io",
+        access_key="Q3AM3UQ867SPQQA43P2F",
+        secret_key="zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG",
+    )
+
+    print("MinIO Python SDK client initialized", client)
```
feorlen, line 50: Is there something nicer this could print instead of the object?

README.md
```diff
@@ -1,69 +1,115 @@
...
+import os
+from minio import Minio
+from minio.error import S3Error
+
 def main():
     # Create a client with the MinIO server playground, its access key
     # and secret key.
     client = Minio(
-        "play.min.io",
+        endpoint="play.min.io",
```
feorlen, line 87: Adding `endpoint=` here follows the same form as the access and secret keys. It works, is this advisable?

README.md
```diff
@@ -1,69 +1,115 @@
...
+
+This example does the following:
+
+- Connects to the MinIO `play` server using the provided credentials.
+- Creates a bucket named `minio-python-sdk-test-bucket`.
+- Uploads a file named `minio-python-sdk-test-file.bin` from `/tmp`.
+- Verifies the file was created using `mc ls`.
+
+```py
+# file_uploader.py MinIO Python SDK example
```
feorlen, line 77: Code feedback welcome. 

Attempted to be slightly more friendly for Windows users, but without fully getting into the cross-platform issues. Mainly by using `os.path.join()` instead of a string path, although a Windows user would still need to update `/tmp` to something appropriate.

I don't have a good way to actually test on Windows.

README.md
```diff
@@ -74,23 +120,55 @@ if __name__ == "__main__":
...
 * [Complete Documentation](https://min.io/docs/minio/kubernetes/upstream/index.html)
 
 ## Contribute
-Please refer [Contributors Guide](https://github.com/minio/minio-py/blob/master/CONTRIBUTING.md)
+
+[Contributors Guide](https://github.com/minio/minio-py/blob/master/CONTRIBUTING.md)
+
+## License
+
+This SDK is distributed under the [Apache License, Version 2.0](https://www.apache.org/licenses/LICENSE-2.0), see [LICENSE](https://github.com/minio/minio-py/blob/master/LICENSE) for more information.
```
feorlen, line 171: The Go one also links to its `NOTICE` but this repo doesn't have that file. Recommendations?

README.md
```diff
@@ -1,69 +1,115 @@
-# MinIO Python SDK for Amazon S3 Compatible Cloud Storage [![Slack](https://slack.min.io/slack?type=svg)](https://slack.min.io)
+# MinIO Python Client SDK for Amazon S3 Compatible Cloud Storage [![Slack](https://slack.min.io/slack?type=svg)](https://slack.min.io) [![Sourcegraph](https://sourcegraph.com/github.com/minio/minio-py/-/badge.svg)](https://sourcegraph.com/github.com/minio/minio-py?badge) [![Apache V2 License](https://img.shields.io/badge/license-Apache%20V2-blue.svg)](https://github.com/minio/minio-py/blob/master/LICENSE)
 
-MinIO Python SDK is Simple Storage Service (aka S3) client to perform bucket and object operations to any Amazon S3 compatible object storage service.
+The MinIO Python Client SDK provides straightforward APIs to access any Amazon S3 compatible object storage.
```
balamurugana, line 3: Technically we provide more higher level API methods (uses lower level APIs used to accomplish specific purpose e.g. put_object, compose_object API) than `straightforward` (low level) APIs. These low level APIs are accessible by inheriting `MinioClient` class.

README.md
```diff
@@ -1,69 +1,115 @@
...
 
-For a complete list of APIs and examples, please take a look at the [Python Client API Reference](https://min.io/docs/minio/linux/developers/python/API.html)
+This Quickstart Guide covers how to install the MinIO client SDK, connect to MinIO, and create a sample file uploader.
+For a complete list of APIs and examples, see the [Python Client API Reference](https://min.io/docs/minio/linux/developers/python/API.html)
 
-## Minimum Requirements
-Python 3.7 or higher.
+These examples presume a working [3.7+ Python development environment](https://www.python.org/downloads/) and the [MinIO `mc` command line tool](https://min.io/docs/minio/linux/reference/minio-mc.html).
```
balamurugana, line 8: This text needs to be placed with the example.

README.md
balamurugana: The prerequisite is user should know what is `pip` and what is python source.

README.md
```diff
@@ -1,69 +1,115 @@
...
+## Install the Minio Python SDK
+
+### Using pip
 
 ```sh
 pip3 install minio
 ```
 
-## Download source
+### From GitHub
```
balamurugana, line 18: `Using source code` is more appropriate.

README.md
balamurugana: This looks odd to me. We would need to provide the example and in the example, we would need to provide meaningful comments in each line to explain what it does.

README.md
```diff
@@ -1,69 +1,115 @@
...
-| Access Key | Access key (aka user ID) of an account in the S3 service.  |
-| Secret Key | Secret key (aka password) of an account in the S3 service. |
+The MinIO client requires the following parameters to connect to an Amazon S3 compatible object storage:
 
-This example uses MinIO server playground [https://play.min.io](https://play.min.io). Feel free to use this service for test and development.
+| Parameter  | Description                                            |
+|------------|--------------------------------------------------------|
+| Endpoint   | URL to S3 service.                                     |
+| Access Key | Access key (user ID) of an account in the S3 service.  |
+| Secret Key | Secret key (password) of an account in the S3 service. |
```
balamurugana, line 34: I would suggest that we would need to remove this and add comment in the example (the current example has that). BTW the parameter names are wrong.

README.md
balamurugana: It doesn't make sense to print `client` object to indicate the object is created successfully. If client creation failed, we get exception in the above line.

README.md
```diff
@@ -1,69 +1,115 @@
...
+    print("MinIO Python SDK client initialized", client)
+
+if __name__ == "__main__":
+    try:
+        main()
+    except S3Error as exc:
+        print("error occurred.", exc)
+```
+
+## Example - File Uploader
```
balamurugana, line 59: I see `minio-go` style README is too verbose for Python. My suggestion is to provide the example directly.

README.md
balamurugana: As MS Windows and Python works with backslash (`/`) to refer file path on the system, I do not think it is necessary to cover MS Windows specifics here.

README.md
balamurugana: It works, but not necessary. The meaning of leaving it as positional argument is to indicate a mandatory argument.

README.md
```diff
@@ -1,69 +1,115 @@
...
     client = Minio(
-        "play.min.io",
+        endpoint="play.min.io",
         access_key="Q3AM3UQ867SPQQA43P2F",
         secret_key="zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG",
     )
 
-    # Make 'asiatrip' bucket if not exist.
-    found = client.bucket_exists("asiatrip")
+    bucket_name = "minio-python-sdk-test-bucket"
```
balamurugana, line 92: These names look like a test suite. As this is an example, we would need to provide a meaningful names for users

README.md
```diff
@@ -74,23 +120,55 @@ if __name__ == "__main__":
...
 
-#### Run File Uploader
+**1. Create a test file containing data:**
+
+You can do this with `dd` on Linux or macOS systems:
+
 ```sh
-$ python file_uploader.py
-'/home/user/Photos/asiaphotos.zip' is successfully uploaded as object 'asiaphotos-2015.zip' to bucket 'asiatrip'.
+dd if=/dev/urandom of=/tmp/minio-python-sdk-test-file.bin bs=2048 count=10
```
balamurugana, line 128: These are all over complicating to run the example. My suggesting is to leave the prerequisite steps to be outside of this README.

README.md
balamurugana: You could copy the file from `minio-go` to here and have a link.

README.md
feorlen: Removed the sourcegraph badge, left the license. If there's something more useful here, can add whatever seems helpful.

README.md
feorlen: After discussing with Docs team, decided to drop specific Windows mentions for now. Included a comment in the example about updating the path/filename, as a prompt for any reader who may need something different.

README.md
feorlen: Corrected the table of args

README.md
feorlen: Reworked example with more generic names

README.md
feorlen: Dropped this for a different approach.

README.md
feorlen: I kept this section, as it seems useful to reinforce needed parameters for less knowledgeable readers. Revised to have a very minimal example of calling the `Minio()` method. 

It's an incomplete code snippet showing the `import`, so maybe there's a more usual way of presenting something like this for Python.

README.md
feorlen: Went back to first positional and others keyword. The table listing the args is now directly above, hopefully will help clarify.

feorlen:
@balamurugana I've reworked the entire page based on our conversation. I've retained some of the descriptive text and streamlined other parts. When you have a moment, I'd appreciate a look and if this approach seems useful for both less and more knowledgeable types of readers. It's closer to the original, but with some additional context for a novice/casual reader.

NOTICE
```diff
@@ -0,0 +1,9 @@
+MinIO Cloud Storage, (C) 2014-2023 MinIO, Inc.
```
feorlen, line 1: Added and linked to this along with the license

README.md
```diff
@@ -1,96 +1,155 @@
...
 
-This example uses MinIO server playground [https://play.min.io](https://play.min.io). Feel free to use this service for test and development.
+For example:
 
-### file_uploader.py
 ```py
 from minio import Minio
-from minio.error import S3Error
 
+client = Minio("play.min.io",
```
feorlen, line 52: Perhaps this is more practical, less verbose yet still some additional context for novice readers.

harshavardhana:
PTAL @balamurugana

minio-trusted:
> PTAL @balamurugana

Bump @balamurugana this is waiting on your review - can you please take a look?

balamurugana (approved):

PR #1327, opened 2023-11-09
Add type annotations to xml.py

trim21:


balamurugana (changes requested):
Duck typing in this PR is incomplete. Please fix accordingly.

trim21:
> Duck typing in this PR is incomplete. Please fix accordingly.

Sorry, which class do you mean? `IFromXML` and `IToXML` are complete?

trim21:
> > > Duck typing in this PR is incomplete. Please fix accordingly.
> > 
> > 
> > Sorry, which class do you mean? `IFromXML` and `IToXML` are incomplete?
> 
> Code in xml.py

done

minio/xml.py
```diff
@@ -80,20 +92,43 @@ def findtext(element, name, strict=False):
...
+    with io.BytesIO() as data:
+        ET.ElementTree(element).write(
+            data,
+            encoding=None,
+            xml_declaration=False,
+        )
+        return data.getvalue()
+
+
+class IToXML(Protocol):
```
balamurugana, line 125: How about using abstract class instead of `Protocol` inheritance?

balamurugana (changes requested):

minio/xml.py
trim21: That would involve many other files

balamurugana: I couldn't able to understand why it involves other files.

trim21: abstract class doesn't work like `Protocol` in typing, abstract class is expected to be inherited as base class, but in typing `Protocol` mean duck typing.

This mean, when a class support `toxml(...)`, this class need to use `IToXMl` as base class, which dould involve other files.

trim21: example:
```python
import abc
from abc import ABC


class ABCToXML(ABC):

    @abc.abstractmethod
    def toxml(self, element: str | None) -> str:
        ...


class ImplCorrect(ABCToXML):
    """"""

    def toxml(self, element: str | None) -> str:
        return ""


class ConsiderNotImplemented:
    """"""

    def toxml(self, element: str | None) -> str:
        return ""


_1: ABCToXML = ImplCorrect()
_2: ABCToXML = ConsiderNotImplemented() # type error here
```

minio/xml.py
```diff
@@ -80,20 +92,43 @@ def findtext(element, name, strict=False):
     return element.text or ""
 
 
-def unmarshal(cls, xmlstring):
+K = TypeVar("K")
```
balamurugana, line 95: Have meaningful name here

minio/xml.py
```diff
@@ -80,20 +92,43 @@ def findtext(element, name, strict=False):
...
+
+
+class IFromXML(Protocol):
```
balamurugana, line 98: Rename this class as `FromXMLType` or something meaningful

minio/xml.py
```diff
@@ -80,20 +92,43 @@ def findtext(element, name, strict=False):
...
+    """typing stub for class with `fromxml` method"""
+
+    @classmethod
+    def fromxml(cls: Type[K], element: ET.Element) -> K:
+        """Create python object with values from XML element."""
+
+
+T = TypeVar("T", bound=IFromXML)
```
balamurugana, line 106: Have a meaningful name here

minio/xml.py
balamurugana, line 125: Have a meaningful name here

balamurugana (changes requested):

minio/xml.py
trim21: this is generic type parameter, it's custom to be named as `T` `K` `J`。

minio/xml.py
```diff
@@ -80,20 +92,43 @@ def findtext(element, name, strict=False):
...
 
-def getbytes(element):
+def getbytes(element: ET.Element) -> bytes:
     """Convert ElementTree.Element to bytes."""
-    data = io.BytesIO()
-    ET.ElementTree(element).write(
-        data, encoding=None, xml_declaration=False,
-    )
-    return data.getvalue()
+    with io.BytesIO() as data:
```
balamurugana, line 116: Why do we need to use `with` here?

balamurugana (changes requested):

minio/xml.py
trim21: close file and discard the buffer.

trim21: https://docs.python.org/3/library/io.html#io.BytesIO

trim21: This is not necessary, gc will handle them if we don't use `with`

balamurugana: Yep. For this function, it is unnecessary.

balamurugana (approved):

minio/xml.py
```diff
@@ -16,28 +16,36 @@
...
+) -> ET.Element:
     """Create ElementTree.Element with tag and namespace."""
-    return ET.Element(tag, {'xmlns': namespace} if namespace else {})
+    return ET.Element(tag, {"xmlns": namespace} if namespace else {})
 
 
-def SubElement(parent, tag, text=None):  # pylint: disable=invalid-name
+def SubElement(  # pylint: disable=invalid-name
+    parent: ET.Element, tag: str, text: str | None = None
+) -> ET.Element:
```
balamurugana, line 40: It returns `ET.SubElement`

balamurugana (changes requested):

minio/xml.py
trim21, line 40:
```suggestion
) -> ET.SubElement:
```

balamurugana (approved):

PR #1334, opened 2023-11-17
Add type hints for MinioAdmin class

jessebot:
ran `make check` and it looks like it is good to go, however, I'm still not sure if credentials should be type Provider or not? Open to feedback to improve this!

### DIFF
--- a/minio/minioadmin.py
+++ b/minio/minioadmin.py
@@ -92,7 +92,7 @@ class MinioAdmin:
                  region: str = "",
                  secure: bool = True,
                  cert_check: bool = True,
-                 http_client: str = None):
+                 http_client: urllib3.poolmanager.PoolManager = None):
         url = _parse_url(("https://" if secure else "http://") + endpoint)
         if not isinstance(credentials, Provider):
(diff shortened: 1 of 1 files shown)


balamurugana (approved):

minio/minioadmin.py
```diff
@@ -86,12 +86,13 @@
...
-                 secure=True,
-                 cert_check=True,
-                 http_client=None):
+    def __init__(self,
+                 endpoint: str,
+                 credentials: Provider,
+                 region: str = "",
+                 secure: bool = True,
+                 cert_check: bool = True,
+                 http_client: str = None):
```
jessebot, line 95:
```suggestion
                 http_client: urllib3.poolmanager.PoolManager = None):
```
This might be more correct. I'm not sure as I haven't used this feature yet.

balamurugana: Yep. This is correct. Fix the PR accordingly.

balamurugana (changes requested):

balamurugana (approved):

PR #1342, opened 2023-11-23
Decrypt message above 16k

Dzordzu:
## Current behavior
As of now decrypting in `crypto.py` works only for small messages. This does affect an entire `minioadmin` module.   

## Introduced solution
Partition the message into smaller chunks. Decode every chunk separately

## Related issues
- [x] #1343

### DIFF
--- a/minio/crypto.py
+++ b/minio/crypto.py
@@ -20,16 +20,52 @@
 """Cryptography to read and write encrypted MinIO Admin payload"""
 
 import os
+import typing
 
 from argon2.low_level import Type, hash_secret_raw
 from Crypto.Cipher import AES, ChaCha20_Poly1305
 
-_NONCE_LEN = 8
+# Encrypted message format:
+#  header | encrypted data
+#   41       ~ len(data)
+#
+# Encrypted data:
+# payload (16 KB) | tag (16 bytes)
+# payload (16 KB) | tag (16 bytes)
+# payload (16 KB) | tag (16 bytes)
+# ...
+# payload (1 byte - 16 KB) | tag (16 bytes)
+#
+# Header format:
+# salt | AEAD ID | nonce
+#  32      1         8
+#
+# Notes: additional data for the cipher algo (wihin every single encrypted
+#   data payload) needs to contain 0x00 byte EXCEPT the last one. The last
+#   is noted by 0x80 byte
+#
+# Important links:
+#  * https://github.com/minio/madmin-go/blob/main/encrypt.go#L43-L49
+#  * https://github.com/secure-io/sio-go/blob/master/sio.go#L25
+#  * https://github.com/secure-io/sio/blob/master/img/channel_construction.svg
+#  * https://github.com/secure-io/sio-go/tree/master
+
+_STD_PACKAGE_SIZE = 1 << 14  # 16KB
 _SALT_LEN = 32
+_NONCE_LEN = 8
+_TAG_LEN = 16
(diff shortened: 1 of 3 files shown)


minio/crypto.py
```diff
@@ -20,16 +20,46 @@
 """Cryptography to read and write encrypted MinIO Admin payload"""
 
 import os
+import typing
 
 from argon2.low_level import Type, hash_secret_raw
 from Crypto.Cipher import AES, ChaCha20_Poly1305
 
-_NONCE_LEN = 8
+# Encrypted message format:
```
balamurugana, line 28: Where do you get this information?

balamurugana (changes requested):

minio/crypto.py
Dzordzu: From [your documentation](https://github.com/minio/madmin-go/blob/main/encrypt.go#L43-L49)

Dzordzu: Also it was present in the [current code](https://github.com/minio/minio-py/blob/master/minio/crypto.py#L27)

Dzordzu: And information that is present in the later part is present in the sio implementation:
* https://github.com/secure-io/sio-go/blob/master/sio.go#L25
* https://github.com/secure-io/sio/blob/master/img/channel_construction.svg
* https://github.com/secure-io/sio-go/tree/master

balamurugana: 1. Please point https://github.com/minio/madmin-go/blob/main/encrypt.go#L43-L49 instead of repeating it here.
2. I don't see more than 16k implementation in `madmin-go`. Could you point why do we need that?

balamurugana (changes requested):

minio/crypto.py
Dzordzu: If you want to reproduce the issue try to create above 3k buckets with members (with long names). Then try to call an endpoint responsible for listing entities with their affiliation

balamurugana: > If you want to reproduce the issue try to create above 3k buckets with members (with long names). Then try to call an endpoint responsible for listing entities with their affiliation

For this scenario, what API have you used to reproduce the issue?

Dzordzu: For example (`minioadmin`, so prefixed with proper path):
```
idp/ldap/policy-entities
idp/builtin/policy-entities
```

`list-users` for bigger instances…

Dzordzu: Anything with potentially more excessively capacious response

Dzordzu:
Squashed commits

minio/crypto.py
```diff
@@ -77,59 +117,106 @@ def encrypt(payload: bytes, password: str) -> bytes:
...
+) -> typing.Union[AesGcmCipherProvider, ChaCha20Poly1305CipherProvider, None]:
+    """Generate cipher basing on the version byte"""
+    if cipher_id == 0:
+        return AesGcmCipherProvider()
+    if cipher_id == 1:
+        return ChaCha20Poly1305CipherProvider()
+    return None
+
+
 def decrypt(payload: bytes, password: str) -> bytes:
```
balamurugana, line 159: We would need to have reader interface than handling `bytes` here.

I am working on this including size > 16KiB for encrypt and decrypt.

balamurugana:
Please refer https://github.com/minio/minio-py/pull/1345. Feel free to test it and close this PR accordingly

PR #1407, opened 2024-03-20
files added

Sajiyah-Salat:
hello, I have made a change


harshavardhana:
The PR is not correct, its not fixing anything here.  The correct word is `uses` not `cases`

Sajiyah-Salat:
okay, I understood @harshavardhana Thanks for the reply.

PR #1429, opened 2024-07-16
Support for operations on Tencent COS

lentitude2tk:


harshavardhana:
We have stopped adding new endpoints in this manner as native endpoints.

There are no reasons to have this.

lentitude2tk:
> We have stopped adding new endpoints in this manner as native endpoints.
> 
> There are no reasons to have this.

Currently, the minio-java-sdk provides the enableVirtualStyleEndpoint method, and the milvus-go-sdk offers the BucketLookup=BucketLookupDNS method, both of which can be dynamically configured externally to support this. However, the minio-python-sdk has not yet exposed this capability externally. Are there no plans to align this functionality with the java and go SDKs in the future?

lentitude2tk:
@balamurugana 
This is my mistake; I didn't notice this method. I can indeed use the enable_virtual_style_endpoint. However, I observed that this setting is not the highest priority. When constructing the BaseUrl, minio enforces the use of pathStyle based on certain parameters' interface behavior. Is there an existing method to avoid this?

https://github.com/minio/minio-py/blob/ffe9045fc8bdb435c02054336c79670ed61704a1/minio/helpers.py#L733

harshavardhana:
@lentitude2tk this the part that needs to be fixed, feel free to send a PR

lentitude2tk:
> @lentitude2tk this the part that needs to be fixed, feel free to send a PR

Sure, thank you for your reply.

lentitude2tk:
> @lentitude2tk this the part that needs to be fixed, feel free to send a PR

I have resubmitted a PR to remove the invalid PathStyle setting https://github.com/minio/minio-py/pull/1430. I reviewed the code of minio-java-sdk, which also includes this check, but it doesn't actually pass the queryParameter, so this part of the logic doesn't work. However, it doesn't affect functionality.

https://github.com/minio/minio-java/blob/245f63d385cc9c58c75696da85eaf19933746d6c/api/src/main/java/io/minio/MinioAsyncClient.java#L1389

PR #1430, opened 2024-07-16
GetBucketLocation API does not require path-style addressing in AWS S3

lentitude2tk:
From the AWS S3 API documentation, we can see that the GetBucketLocation API supports virtual style access.Tested locally without any issues.

https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html

image


balamurugana (changes requested):
As `GetBucketLocation` is deprecated, actually we suppose to use  `HeadBucket` to fetch location instead of `GetBucketLocation`. I would like to leave it as is ATM.

lentitude2tk:
> As `GetBucketLocation` is deprecated, actually we suppose to use `HeadBucket` to fetch location instead of `GetBucketLocation`. I would like to leave it as is ATM.

From the perspective of code referencing hierarchy, almost all interface methods in minio-py are based on the GetBucketLocation method, first querying the region based on the bucket, and then performing secondary operations. Does the activation of GetBucketLocation refer to the Minio layer or the AWS side?

https://github.com/minio/minio-py/blob/ffe9045fc8bdb435c02054336c79670ed61704a1/minio/api.py#L437

lentitude2tk:
HeadBucket also relies on the processing of the `self._get_region(bucket_name)` method, which itself uses the `GetBucketLocation` method.

balamurugana:
It is safer side to keep using path style for `GetBucketLocation` for many object storage services not limiting to AWS S3.

PR #1433, opened 2024-08-14
fix endpoint compatibility

airbot-zn:
supported endpoint=http://servicename .


balamurugana (changes requested):
We don't need to support this format.

Issue #1411, opened 2024-04-10
IMDS V2 support for AWS


ShuaiShao93:
It seems we don't support [IMDS V2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) endpoint in [IamAwsProvider](https://github.com/minio/minio-py/blob/master/minio/credentials/providers.py#L455-L465). Is it possible to support it?

balamurugana:
@ShuaiShao93 feel free to send a PR

setu4993:
I'll have a PR for this in the next couple of days.

PR #1437, opened 2024-08-20
update IamAwsProvider as per minio-go implementation

setu4993:
This PR updates the `IamAwsProvider` to use IMDSv2, which is [automatically enabled](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-IMDS-new-instances.html) for all AWS EC2 / ECS instances.

I'm using this on an internal project successfully since ~a week, so contributing it back upstream.

Closes https://github.com/minio/minio-py/issues/1411.

### DIFF
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -32,7 +32,6 @@ from datetime import timedelta
 from pathlib import Path
 from typing import Callable, cast
 from urllib.parse import urlencode, urlsplit, urlunsplit
-
 from xml.etree import ElementTree as ET
 
 import certifi
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -402,10 +402,9 @@ class IamAwsProvider(Provider):
         self._full_uri = os.environ.get("AWS_CONTAINER_CREDENTIALS_FULL_URI")
         self._credentials: Credentials | None = None
 
-    def fetch(self, url: str) -> Credentials:
-        """Fetch credentials from EC2/ECS. """
-
-        res = _urlopen(self._http_client, "GET", url)
+    def fetch(self, url: str, headers: dict) -> Credentials:
+        """Fetch credentials from EC2/ECS."""
+        res = _urlopen(self._http_client, "GET", url, headers=headers)
         data = json.loads(res.data)
         if data.get("Code", "Success") != "Success":
             raise ValueError(
@@ -457,17 +456,27 @@ class IamAwsProvider(Provider):
                     "http://169.254.169.254" +
                     "/latest/meta-data/iam/security-credentials/"
                 )
-
-            res = _urlopen(self._http_client, "GET", url)
+            # Step 1 of the IMDSv2 protocol: get a token from the metadata
+            # service with a 6-hour TTL.
+            response = _urlopen(
+                self._http_client, "PUT",
+                "http://169.254.169.254/latest/api/token",
+                headers={"X-aws-ec2-metadata-token-ttl-seconds": "21600"}
+            )
+            # Step 2: get the role name from the metadata service, with the
+            # token as a header.
+            token_header: dict = {
+                "X-aws-ec2-metadata-token":
+                response.data.decode("utf-8").strip()
+            }
+            res = _urlopen(self._http_client, "GET", url, headers=token_header)
             role_names = res.data.decode("utf-8").split("\n")
             if not role_names:
                 raise ValueError(f"no IAM roles attached to EC2 service {url}")
             url += "/" + role_names[0].strip("\r")
-
         if not url:
             raise ValueError("url is empty; this should not happen")
-
-        self._credentials = self.fetch(url)
+        self._credentials = self.fetch(url, headers=token_header)
         return self._credentials
 
 
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -419,7 +419,7 @@ class IamAwsProvider(Provider):
             relative_uri
         )
         if self._relative_uri and not self._relative_uri.startswith("/"):
-             self._relative_uri = "/" + self._relative_uri
+            self._relative_uri = "/" + self._relative_uri
         self._full_uri = (
             os.environ.get("AWS_CONTAINER_CREDENTIALS_FULL_URI") or
             full_uri
@@ -496,7 +496,7 @@ class IamAwsProvider(Provider):
             # Get IMDS Token
             res = _urlopen(
                 self._http_client,
-                "GET",
+                "PUT",
                 url+"/latest/api/token",
                 headers={"X-aws-ec2-metadata-token-ttl-seconds": "21600"},
             )
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -29,7 +29,8 @@ from abc import ABCMeta, abstractmethod
 from datetime import timedelta
 from pathlib import Path
 from typing import Callable, cast
-from urllib.parse import urlencode, urlsplit
+from urllib.parse import urlencode, urlsplit, urlunsplit
+
 from xml.etree import ElementTree as ET
 
 import certifi
@@ -42,7 +43,7 @@ except ImportError:
 
 from urllib3.util import Retry, parse_url
 
-from minio.helpers import sha256_hash
+from minio.helpers import sha256_hash, url_replace
 from minio.signer import sign_v4_sts
 from minio.time import from_iso8601utc, to_amz_date, utcnow
 from minio.xml import find, findtext
@@ -381,6 +382,13 @@ class IamAwsProvider(Provider):
             self,
             custom_endpoint: str | None = None,
             http_client: PoolManager | None = None,
+            auth_token: str | None = None,
+            relative_uri: str | None = None,
+            full_uri: str | None = None,
+            token_file: str | None = None,
+            role_arn: str | None = None,
+            role_session_name: str | None = None,
+            region: str | None = None,
     ):
         self._custom_endpoint = custom_endpoint
         self._http_client = http_client or PoolManager(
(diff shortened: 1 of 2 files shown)


minio/credentials/providers.py
```diff
@@ -402,10 +402,9 @@ def __init__(
         self._full_uri = os.environ.get("AWS_CONTAINER_CREDENTIALS_FULL_URI")
         self._credentials: Credentials | None = None
 
-    def fetch(self, url: str) -> Credentials:
-        """Fetch credentials from EC2/ECS. """
-
-        res = _urlopen(self._http_client, "GET", url)
+    def fetch(self, url: str, headers: dict) -> Credentials:
```
balamurugana, line 405: We would need to sync with https://github.com/minio/minio-go/blob/master/pkg/credentials/iam_aws.go.

setu4993: Thanks, @balamurugana . Can you please elaborate what that looks like? First time contributor here and didn't see a note about it in the contributing doc.

balamurugana: I am saying we should use the same logic as in `minio-go`. Below is the actual implementation
```diff
diff --git a/minio/credentials/providers.py b/minio/credentials/providers.py
index 51dca1f..8790054 100644
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# pylint: disable=too-many-branches
+
 """Credential providers."""
 
 from __future__ import annotations
@@ -29,7 +31,7 @@ from abc import ABCMeta, abstractmethod
 from datetime import timedelta
 from pathlib import Path
 from typing import Callable, cast
-from urllib.parse import urlencode, urlsplit
+from urllib.parse import urlencode, urlsplit, urlunsplit
 from xml.etree import ElementTree as ET
 
 import certifi
@@ -42,7 +44,7 @@ except ImportError:
 
 from urllib3.util import Retry, parse_url
 
-from minio.helpers import sha256_hash
+from minio.helpers import sha256_hash, url_replace
 from minio.signer import sign_v4_sts
 from minio.time import from_iso8601utc, to_amz_date, utcnow
 from minio.xml import find, findtext
@@ -381,6 +383,13 @@ class IamAwsProvider(Provider):
             self,
             custom_endpoint: str | None = None,
             http_client: PoolManager | None = None,
+            auth_token: str | None = None,
+            relative_uri: str | None = None,
+            full_uri: str | None = None,
+            token_file: str | None = None,
+            role_arn: str | None = None,
+            role_session_name: str | None = None,
+            region: str | None = None,
     ):
         self._custom_endpoint = custom_endpoint
         self._http_client = http_client or PoolManager(
@@ -390,22 +399,42 @@ class IamAwsProvider(Provider):
                 status_forcelist=[500, 502, 503, 504],
             ),
         )
-        self._token_file = os.environ.get("AWS_WEB_IDENTITY_TOKEN_FILE")
-        self._aws_region = os.environ.get("AWS_REGION")
-        self._role_arn = os.environ.get("AWS_ROLE_ARN")
-        self._role_session_name = os.environ.get("AWS_ROLE_SESSION_NAME")
-        self._relative_uri = os.environ.get(
-            "AWS_CONTAINER_CREDENTIALS_RELATIVE_URI",
+        self._token = (
+            os.environ.get("AWS_CONTAINER_AUTHORIZATION_TOKEN") or
+            auth_token
+        )
+        self._token_file = (
+            os.environ.get("AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE") or
+            auth_token
+        )
+        self._identity_file = (
+            os.environ.get("AWS_WEB_IDENTITY_TOKEN_FILE") or token_file
+        )
+        self._aws_region = os.environ.get("AWS_REGION") or region
+        self._role_arn = os.environ.get("AWS_ROLE_ARN") or role_arn
+        self._role_session_name = (
+            os.environ.get("AWS_ROLE_SESSION_NAME") or role_session_name
+        )
+        self._relative_uri = (
+            os.environ.get("AWS_CONTAINER_CREDENTIALS_RELATIVE_URI") or
+            relative_uri
         )
         if self._relative_uri and not self._relative_uri.startswith("/"):
             self._relative_uri = "/" + self._relative_uri
-        self._full_uri = os.environ.get("AWS_CONTAINER_CREDENTIALS_FULL_URI")
+        self._full_uri = (
+            os.environ.get("AWS_CONTAINER_CREDENTIALS_FULL_URI") or
+            full_uri
+        )
         self._credentials: Credentials | None = None
 
-    def fetch(self, url: str) -> Credentials:
+    def fetch(
+            self,
+            url: str,
+            headers: dict[str, str | list[str] | tuple[str]] | None = None,
+    ) -> Credentials:
         """Fetch credentials from EC2/ECS. """
 
-        res = _urlopen(self._http_client, "GET", url)
+        res = _urlopen(self._http_client, "GET", url, headers=headers)
         data = json.loads(res.data)
         if data.get("Code", "Success") != "Success":
             raise ValueError(
@@ -428,14 +457,17 @@ class IamAwsProvider(Provider):
             return self._credentials
 
         url = self._custom_endpoint
-        if self._token_file:
+
+        if self._identity_file:
             if not url:
                 url = "https://sts.amazonaws.com"
                 if self._aws_region:
                     url = f"https://sts.{self._aws_region}.amazonaws.com"
+                    if self._aws_region.startswith("cn-"):
+                        url += ".cn"
 
             provider = WebIdentityProvider(
-                lambda: _get_jwt_token(cast(str, self._token_file)),
+                lambda: _get_jwt_token(cast(str, self._identity_file)),
                 url,
                 role_arn=self._role_arn,
                 role_session_name=self._role_session_name,
@@ -444,21 +476,48 @@ class IamAwsProvider(Provider):
             self._credentials = provider.retrieve()
             return cast(Credentials, self._credentials)
 
+        headers: dict[str, str | list[str] | tuple[str]] | None = None
         if self._relative_uri:
             if not url:
                 url = "http://169.254.170.2" + self._relative_uri
+            headers = {"Authorization": self._token} if self._token else None
         elif self._full_uri:
-            if not url:
+            token = self._token
+            if self._token_file:
                 url = self._full_uri
-            _check_loopback_host(url)
+                with open(self._token_file, encoding="utf-8") as file:
+                    token = file.read()
+            else:
+                if not url:
+                    url = self._full_uri
+                    _check_loopback_host(url)
+            headers = {"Authorization": token} if token else None
         else:
             if not url:
-                url = (
-                    "http://169.254.169.254" +
-                    "/latest/meta-data/iam/security-credentials/"
-                )
-
-            res = _urlopen(self._http_client, "GET", url)
+                url = "http://169.254.169.254"
+
+            # Get IMDS Token
+            res = _urlopen(
+                self._http_client,
+                "GET",
+                url+"/latest/api/token",
+                headers={"X-aws-ec2-metadata-token-ttl-seconds": "21600"},
+            )
+            token = res.data.decode("utf-8")
+            headers = {"X-aws-ec2-metadata-token": token} if token else None
+
+            # Get role name
+            res = _urlopen(
+                self._http_client,
+                "GET",
+                urlunsplit(
+                    url_replace(
+                        urlsplit(url),
+                        path="/latest/meta-data/iam/security-credentials/",
+                    ),
+                ),
+                headers=headers,
+            )
             role_names = res.data.decode("utf-8").split("\n")
             if not role_names:
                 raise ValueError(f"no IAM roles attached to EC2 service {url}")
@@ -467,7 +526,7 @@ class IamAwsProvider(Provider):
         if not url:
             raise ValueError("url is empty; this should not happen")
 
-        self._credentials = self.fetch(url)
+        self._credentials = self.fetch(url, headers)
         return self._credentials
 
 
```

balamurugana (changes requested):

minio/credentials/providers.py
```diff
@@ -444,30 +474,55 @@ def retrieve(self) -> Credentials:
...
-                    "/latest/meta-data/iam/security-credentials/"
-                )
-
-            res = _urlopen(self._http_client, "GET", url)
+                url = "http://169.254.169.254"
+
+            # Get IMDS Token
+            res = _urlopen(
+                self._http_client,
+                "PUT",
```
setu4993, line 500: @balamurugana : This line differs from your suggestion because this needs to be a `PUT` request, not `GET`.

`minio-go` also has this as a `PUT` [here](https://github.com/minio/minio-go/blob/e337e77c06bbf8a65247fb018e59c1263e4d86e1/pkg/credentials/iam_aws.go#L334).

minio/credentials/providers.py
setu4993: Wow, thanks for the suggestion and helping me finish the PR :).

I didn't realize this package had drifted so far away from the Go implementation.

setu4993:
Just pushed up the updates.

tests/unit/credentials_test.py
```diff
@@ -45,11 +45,16 @@ def test_credentials_get(self):
         self.assertEqual(creds.session_token, None)
 
 
-class CredListResponse(object):
+class TokenResponse(object):
```
balamurugana, line 48: Remove changes in this unit test which doesn't add any value

setu4993: I think all changes here add value. Status and the data here are being used in the first request's response. And we can see it be used.

I suppose we could use the same response for `TokenResponse` and `CredList` but I do think they should be different because they are different interactions.

balamurugana: As this test (not your change) doesn't prevent any logical error in the code, there is no point updating it. The right way to test is to run with AWS.

balamurugana (changes requested):

tests/unit/credentials_test.py
setu4993: I agree that it doesn't have coverage for all of the cases, but the addition does atleast illustrate the pathway for the IMDS route. I can add a comment to the test to make it explicit that it is only testing the happy path, if that'd help?

balamurugana: The purpose of unit test is to cover all possible values. That's the reason I do not like such partial tests.

setu4993: Ok, I removed the test for `IamAwsProvider` in https://github.com/minio/minio-py/pull/1437/commits/3a5cd228e85c9c44b9313ef72ef2bc865687aff9.

setu4993: FWIW, I did test it with AWS (atleast the original branch that I had) and that worked well. I can verify the exact code here tomorrow, too.

balamurugana (approved):

harshavardhana (approved):

PR #1439, opened 2024-08-26
[Fix] Not splitting object names with type of bytes

BinDruid:
In release **7.2.8** utility method `check_non_empty_string` was replaced with `check_object_name`.

Signature of old method requires the argument `object_name` to be a union of types `str` or `bytes`.

However the newer version `check_object_name` always assume `object_name` is of type `str`.

This assumption causes a compatibility issue for example with `django-minio-backend` package which passes `object_name` with type `bytes`, raising `TypeError` exception.

### DIFF
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -267,10 +267,10 @@ def check_non_empty_string(string: str | bytes):
         raise TypeError() from exc
 
 
-def check_object_name(object_name: str):
+def check_object_name(object_name: str | bytes):
     """Check whether given object name is valid."""
     check_non_empty_string(object_name)
-    tokens = object_name.split("/")
+    tokens = (object_name.decode() if isinstance(object_name, bytes) else object_name).split("/")
     if "." in tokens or ".." in tokens:
         raise ValueError(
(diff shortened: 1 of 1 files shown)


balamurugana (changes requested):
Object name has to be a string as per AWS S3 specification.

BinDruid:
Given the source code of `django-minio-backend` which eventually passes `object_name` and calls `check_object_name` :

```python
  try:
      u: str = client.presigned_get_object(
          bucket_name=self.bucket,
          object_name=name.encode('utf-8'),
          expires=get_setting("MINIO_URL_EXPIRY_HOURS", timedelta(days=7))  # Default is 7 days
      )
      return u
  except urllib3.exceptions.MaxRetryError:
      raise ConnectionError("Couldn't connect to Minio. Check django_minio_backend parameters in Django-Settings")

```

The value of `object-name` is always encoded and passed as bytes not string.

Given the AWS S3 specification, I assume this a compatibility issue not for `minio` but for `django-minio-backend`?

balamurugana:
@BinDruid You would need to fix `django-minio-backend`

PR #1453, opened 2024-10-28
remove EOLed python 3.8 support

balamurugana:


harshavardhana (approved):

Emkas:
Does it mean that Python 3.9 is a minimal version? Readme still tells us about 3.7+

PR #832, opened 2019-12-15
Black, Isort, Pylint

chickeaterbanana:
If the library is formatted with black and isort, it would be easier to contribute. This Pull-Request lets pylint return with a score from 10/10.

Issue #826, opened 2019-11-27
Support for Amazon S3 Transfer Acceleration?


cosmosgenius:
Minio-go already has support for it. Is there any plan to incorporate the same in python binding?

harshavardhana:
Definitely we can add, feel free to send a PR @cosmosgenius

cosmosgenius:
Hey i have opened a PR(#827) with a rough implementation. Let me know

harshavardhana:
Fixed by https://github.com/minio/minio-py/pull/827

PR #1455, opened 2024-10-30
fix calling cleanup function in Worker.run() to avoid race condition

detailyang:
Hello. 

Found that minio encounters deadlocks when uploading large files in certain cases. Further investigation reveals that the thread pool can experience deadlocks under specific conditions. 

The minimal reproduction code is as follows: 
```python
from threading import BoundedSemaphore, Thread
from queue import Queue
import time

class Worker(Thread):
    def __init__(self, tasks_queue, results_queue, exceptions_queue):
        Thread.__init__(self)
        self.tasks_queue = tasks_queue
        self.results_queue = results_queue
        self.exceptions_queue = exceptions_queue
        self.daemon = True
        self.start()

    def run(self):
        while True:
            task = self.tasks_queue.get()
            if not task:
                self.tasks_queue.task_done()
                break
            
            if self.exceptions_queue.empty():
                func, args, kargs, cleanup_func = task
                try:
                    result = func(*args, **kargs)
                    self.results_queue.put(result)
                except Exception as ex:
                    print(f"exception occurred: {ex}")
                    self.exceptions_queue.put(ex)
                finally:
                    cleanup_func()
            else:
                print("Skipping task due to previous exception")
            self.tasks_queue.task_done()

class ThreadPool:
    def __init__(self, num_threads):
        self.results_queue = Queue()
        self.exceptions_queue = Queue()
        self.tasks_queue = Queue()
        self.sem = BoundedSemaphore(num_threads)
        self.num_threads = num_threads

    def add_task(self, func, *args, **kargs):
        print(f"Trying to acquire semaphore at {time.strftime('%H:%M:%S')}")
        self.sem.acquire()
        print(f"Semaphore acquired at {time.strftime('%H:%M:%S')}")
        cleanup_func = self.sem.release
        self.tasks_queue.put((func, args, kargs, cleanup_func))

    def start_parallel(self):
        for _ in range(self.num_threads):
            Worker(self.tasks_queue, self.results_queue, self.exceptions_queue)

    def result(self):
        for _ in range(self.num_threads):
            self.tasks_queue.put(None)
        self.tasks_queue.join()
        if not self.exceptions_queue.empty():
            raise self.exceptions_queue.get()
        return self.results_queue

def task_with_exception():
    raise ValueError("Task failed!")

def normal_task():
    time.sleep(1)
    return "Task completed"

if __name__ == "__main__":
    pool = ThreadPool(num_threads=1) 
    pool.start_parallel()

    for i in range(3):
        if i == 0:
            print(f"\nAdding task {i} (will raise exception)...")
            pool.add_task(task_with_exception)
        else:
            print(f"\nAdding task {i} (normal task)...")
            pool.add_task(normal_task)
        time.sleep(0.1) 
    
    try:
        results = pool.result()
        while not results.empty():
            print(results.get())
    except Exception as e:
        print(f"Final exception: {e}")
```

the main reason is that when there are exceptions in the exception queue, the worker does not release the semaphore, which leads to add_task being permanently blocked. 

The fix is quite simple: ensure that the semaphore is released after each task is retrieved.

### DIFF
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -840,7 +840,8 @@ class Worker(Thread):
                 except Exception as ex:  # pylint: disable=broad-except
                     self._exceptions_queue.put(ex)
 
-            # call cleanup i.e. Semaphore.release irrespective of task execution to avoid race condition.
+            # call cleanup i.e. Semaphore.release irrespective of task
+            # execution to avoid race condition.
             cleanup_func()
             # Mark this task as done, whether an exception happened or not
             self._tasks_queue.task_done()
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -839,8 +839,10 @@ class Worker(Thread):
                     self._results_queue.put(result)
                 except Exception as ex:  # pylint: disable=broad-except
                     self._exceptions_queue.put(ex)
-                finally:
-                    cleanup_func()
+
+            # cleanup_func represents semaphore.release() to manage thread pool capacity,
+            # it should be called unconditionally to avoid semaphore leaks and deadlocks.
+            cleanup_func()
             # Mark this task as done, whether an exception happened or not
             self._tasks_queue.task_done()
 
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -840,8 +840,7 @@ class Worker(Thread):
                 except Exception as ex:  # pylint: disable=broad-except
                     self._exceptions_queue.put(ex)
 
-            # cleanup_func represents semaphore.release() to manage thread pool capacity,
-            # it should be called unconditionally to avoid semaphore leaks and deadlocks.
+            # call cleanup i.e. Semaphore.release irrespective of task execution to avoid race condition.
             cleanup_func()
             # Mark this task as done, whether an exception happened or not
             self._tasks_queue.task_done()
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -829,11 +829,11 @@ class Worker(Thread):
             if not task:
                 self._tasks_queue.task_done()
                 break
+            
+            func, args, kargs, cleanup_func = task
             # No exception detected in any thread,
             # continue the execution.
             if self._exceptions_queue.empty():
-                # Execute the task
-                func, args, kargs, cleanup_func = task
                 try:
                     result = func(*args, **kargs)
                     self._results_queue.put(result)
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -829,7 +829,6 @@ class Worker(Thread):
             if not task:
                 self._tasks_queue.task_done()
                 break
-            
             func, args, kargs, cleanup_func = task
             # No exception detected in any thread,
             # continue the execution.


minio/helpers.py
```diff
@@ -839,8 +839,10 @@ def run(self):
                     self._results_queue.put(result)
                 except Exception as ex:  # pylint: disable=broad-except
                     self._exceptions_queue.put(ex)
-                finally:
-                    cleanup_func()
+
+            # cleanup_func represents semaphore.release() to manage thread pool capacity,
+            # it should be called unconditionally to avoid semaphore leaks and deadlocks.
+            cleanup_func()
```
balamurugana, line 845: I do not understand the fix. How does it different from `finally`. If it is `python` bug, report `python`.

balamurugana (changes requested):

minio/helpers.py
detailyang: see the minimal reproduction code, it would block forever @balamurugana

balamurugana: As the reproducer is generic python code, why is it minio-py issue?

rafsaf: fixing my PR today I lurked to this one out of curiosity, and clearly @detailyang is right

> As the reproducer is generic python code, why is it minio-py issue?

I think because minio-py is using this reproducer in it's code and reproducer has bug :smile: 

---

You can leave "cleanup_func()" in finally block as is, it doesn't matter (but with the line added, it's redundant). The bug is that if **code doesn't go inside** `if self._exceptions_queue.empty()` if statement block, then semaphore is never released (it never reaches finally block)

In other words, it's directly related to using acquire in add_task, and releasing in workers. There is a comment `# Mark this task as done, whether an exception happened or not` and note that marking task as done should also always release semaphore

detailyang: @balamurugana Please take another look at this; the bug is actually quite serious and affects the majority of versions. Many thanks:)

minio/helpers.py
```diff
@@ -839,8 +839,10 @@ def run(self):
                     self._results_queue.put(result)
                 except Exception as ex:  # pylint: disable=broad-except
                     self._exceptions_queue.put(ex)
-                finally:
-                    cleanup_func()
+
+            # cleanup_func represents semaphore.release() to manage thread pool capacity,
+            # it should be called unconditionally to avoid semaphore leaks and deadlocks.
```
balamurugana, line 844:
```suggestion
            # call cleanup i.e. Semaphore.release irrespective of task execution to avoid race condition.
```

detailyang:
> Move line #826 before line #832

Hello.  @balamurugana.

I made a subtle adjustment to ensure cleanup_func is extracted from task to avoid used-before-assignment `cleanup_func`

minio/helpers.py
```diff
@@ -829,18 +829,20 @@ def run(self):
             if not task:
                 self._tasks_queue.task_done()
                 break
+
```
balamurugana, line 832:
```suggestion

```

balamurugana (approved):

harshavardhana (approved):

PR #1460, opened 2024-11-20
Fix Addresses Used in IamAwsProvider

### LINKED ISSUE #1459: Invalid Address Used for IMDSv2
When using `IamAwsProvider` to retrieve credentials, the address used to fetch the session token is invalid, resulting in an error.

## Steps to reproduce
Run the following code snippet in an EC2 instance:
```python
from minio.credentials.providers import IamAwsProvider
IamAwsProvider().retrieve()
```
### Expected Behavior
The credentials should be retrieved successfully without any errors.
### Actual Behavior
An exception is raised:
```
ValueError: http://169.254.169.254/role-name failed with HTTP status code 404
```
## Additional Context
- EC2 instance has IMDSv2 set to Required (IMDSv1 is disabled)
- Versions: `master`, `7.2.11`
- Related to #1437

sam-mosleh:
Fixes #1459

### DIFF
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -508,18 +508,13 @@ class IamAwsProvider(Provider):
             res = _urlopen(
                 self._http_client,
                 "GET",
-                urlunsplit(
-                    url_replace(
-                        urlsplit(url),
-                        path="/latest/meta-data/iam/security-credentials/",
-                    ),
-                ),
+                url+"/latest/meta-data/iam/security-credentials/",
                 headers=headers,
             )
             role_names = res.data.decode("utf-8").split("\n")
             if not role_names:
                 raise ValueError(f"no IAM roles attached to EC2 service {url}")
-            url += "/" + role_names[0].strip("\r")
+            url += "/latest/meta-data/iam/security-credentials/" + role_names[0].strip("\r")
         if not url:
             raise ValueError("url is empty; this should not happen")
         self._credentials = self.fetch(url, headers=headers)
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -31,7 +31,7 @@ from abc import ABCMeta, abstractmethod
 from datetime import timedelta
 from pathlib import Path
 from typing import Callable, cast
-from urllib.parse import urlencode, urlsplit, urlunsplit
+from urllib.parse import urlencode, urlsplit
 from xml.etree import ElementTree as ET
 
 import certifi
@@ -44,7 +44,7 @@ except ImportError:
 
 from urllib3.util import Retry, parse_url
 
-from minio.helpers import sha256_hash, url_replace
+from minio.helpers import sha256_hash
 from minio.signer import sign_v4_sts
 from minio.time import from_iso8601utc, to_amz_date, utcnow
 from minio.xml import find, findtext
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -514,7 +514,8 @@ class IamAwsProvider(Provider):
             role_names = res.data.decode("utf-8").split("\n")
             if not role_names:
                 raise ValueError(f"no IAM roles attached to EC2 service {url}")
-            url += "/latest/meta-data/iam/security-credentials/" + role_names[0].strip("\r")
+            url += "/latest/meta-data/iam/security-credentials/" + \
+                role_names[0].strip("\r")
         if not url:
             raise ValueError("url is empty; this should not happen")
         self._credentials = self.fetch(url, headers=headers)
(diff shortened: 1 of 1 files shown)


minio/credentials/providers.py
```diff
@@ -508,18 +508,13 @@ def retrieve(self) -> Credentials:
             res = _urlopen(
                 self._http_client,
                 "GET",
-                urlunsplit(
-                    url_replace(
-                        urlsplit(url),
-                        path="/latest/meta-data/iam/security-credentials/",
-                    ),
-                ),
+                url+"/latest/meta-data/iam/security-credentials/",
```
balamurugana, line 511: This breaks if `custom_endpoint` is passed. I think you would need to leave it as is.

minio/credentials/providers.py
```diff
@@ -508,18 +508,13 @@ def retrieve(self) -> Credentials:
...
                 headers=headers,
             )
             role_names = res.data.decode("utf-8").split("\n")
             if not role_names:
                 raise ValueError(f"no IAM roles attached to EC2 service {url}")
-            url += "/" + role_names[0].strip("\r")
+            url += "/latest/meta-data/iam/security-credentials/" + role_names[0].strip("\r")
```
balamurugana, line 517: May be you would need to do to same logic as in line #511

balamurugana (changes requested):

minio/credentials/providers.py
sam-axcient: Thanks for the feedback!

I noticed the same logic is already applied on 
https://github.com/minio/minio-py/blob/759a5f3f432e884823d040e1746600c537ad6c41/minio/credentials/providers.py#L501

which follows the same pattern used in the Go implementation 
https://github.com/minio/minio-go/blob/ab1c2fe050a4dd502c863178696a54c44610ca82/pkg/credentials/iam_aws.go#L334

However, the Go implementation has an issue: it doesn’t account for cases where a custom endpoint contains a path. This leads to the assumption that the custom_endpoint doesn’t have a path, which is why I addressed it this way on line 511.

To resolve this, we have two options when a custom_endpoint is passed:

1. **Assume it does not contain a path** and use simple string concatenation (`+`) in all instances where we manipulate the URL.
2. **Assume it can contain a path** and ensure proper handling by using urllib.parse.SplitResult in all usages to safely override the path.

Given that the Go implementation itself is inconsistent—sometimes appending directly to the URL and other times parsing the URL—it would be helpful to align on one approach to avoid further confusion and potential issues.

Which approach do you think would be best to adopt moving forward?

sam-mosleh: Thanks for the feedback!

I noticed the same logic is already applied on 
https://github.com/minio/minio-py/blob/759a5f3f432e884823d040e1746600c537ad6c41/minio/credentials/providers.py#L501

which follows the same pattern used in the Go implementation 
https://github.com/minio/minio-go/blob/ab1c2fe050a4dd502c863178696a54c44610ca82/pkg/credentials/iam_aws.go#L334

However, the Go implementation has an issue: it doesn’t account for cases where a custom endpoint contains a path. This leads to the assumption that the custom_endpoint doesn’t have a path, which is why I addressed it this way on line 511.

To resolve this, we have two options when a custom_endpoint is passed:

1. **Assume it does not contain a path** and use simple string concatenation (`+`) in all instances where we manipulate the URL.
2. **Assume it can contain a path** and ensure proper handling by using urllib.parse.SplitResult in all usages to safely override the path.

Given that the Go implementation itself is inconsistent—sometimes appending directly to the URL and other times parsing the URL—it would be helpful to align on one approach to avoid further confusion and potential issues.

Which approach do you think would be best to adopt moving forward?

sam-mosleh:
@balamurugana I totally get that these commits might get rewritten completely depending on the approach you choose, but my OCD brain just couldn’t resist not fixing these linting errors! 😅

sam-mosleh:
@balamurugana  Since Python 3.7+ includes a built-in `_replace` method which is the preferred way to alter the values, would you be open to refactoring the code to remove the custom `url_replace` function?

If you're on board with this, I can submit a follow-up PR to make the change.

balamurugana:
> @balamurugana Since Python 3.7+ includes a built-in `_replace` method which is the preferred way to alter the `SplitResult` values, would you be open to refactoring the code to remove the custom `url_replace` function?
> 
> If you're on board with this, I can submit a follow-up PR to make the change.

@sam-mosleh You could leave the code as is.

setu4993:
I just noticed this. I just created #1464 which is an alternative implementation for the same fix.

setu4993:
I don't have a preference on which one we move forward with but would love to see either be merged + released soon.

sam-mosleh:
@setu4993 Your implementation of the fix exactly matches the initial implementation in this PR. However, there were changes requested earlier. Thanks for the ping!

@balamurugana Is there anything blocking the merge? It’s been almost two weeks without feedback, and I’d appreciate any updates on this.

balamurugana:
I am closing on behalf of https://github.com/minio/minio-py/pull/1465

PR #1463, opened 2024-12-05
Add account_info Admin API

mhkarimi1383:
Account info admin command was missing in admin client and needed to get bucket information (just like what we have in console)

### DIFF
--- a/minio/minioadmin.py
+++ b/minio/minioadmin.py
@@ -323,11 +323,14 @@ class MinioAdmin:
         )
         return response.data.decode()
 
-    def account_info(self) -> str:
-        """Get MinIO Account (aka User) information."""
+    def account_info(self, prefix_usage: bool = False) -> str:
+        """Get usage information for the authenticating account"""
         response = self._url_open(
             "GET",
             _COMMAND.ACCOUNT_INFO,
+            query_params={
+                "prefix-usage": str(prefix_usage).lower(),
+            },
         )
         return response.data.decode()
 
--- a/minio/minioadmin.py
+++ b/minio/minioadmin.py
@@ -53,6 +53,7 @@ from .signer import sign_v4_s3
 @unique
 class _COMMAND(Enum):
     """Admin Command enumerations."""
+    ACCOUNT_INFO = "accountinfo"
     ADD_USER = "add-user"
     USER_INFO = "user-info"
     LIST_USERS = "list-users"
@@ -313,6 +314,14 @@ class MinioAdmin:
         response = self._url_open(
             "GET",
             _COMMAND.INFO,
+        )
+        return response.data.decode()
+
+    def account_info(self) -> str:
+        """Get MinIO Account (aka User) information."""
+        response = self._url_open(
+            "GET",
+            _COMMAND.ACCOUNT_INFO,
         )
         return response.data.decode()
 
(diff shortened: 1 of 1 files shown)


harshavardhana (dismissed):

minio-trusted:
PTAL @balamurugana

minio/minioadmin.py
```diff
@@ -322,6 +323,14 @@ def info(self) -> str:
         )
         return response.data.decode()
 
+    def account_info(self) -> str:
```
balamurugana, line 326: Refer https://github.com/minio/madmin-go/blob/main/user-commands.go#L80
1. Add comment as per `madmin-go`
2. Add argument `prefix_usage: bool = false` and handle it

balamurugana (changes requested):

minio/minioadmin.py
mhkarimi1383: Fixed.

minio/minioadmin.py
```diff
@@ -322,6 +323,17 @@ def info(self) -> str:
         )
         return response.data.decode()
 
+    def account_info(self, prefix_usage: bool = False) -> str:
+        """Get usage information for the authenticating account"""
+        response = self._url_open(
+            "GET",
+            _COMMAND.ACCOUNT_INFO,
+            query_params={
```
balamurugana, line 331: Send `query_params` only if `prefix_usage` is `true`

balamurugana (changes requested):

minio/minioadmin.py
mhkarimi1383: Fixed, Also switched to Mapping in _url_open for query params to avoid further type errors

minio/minioadmin.py
```diff
@@ -163,7 +164,7 @@ def _url_open(
             self,
             method: str,
             command: _COMMAND,
-            query_params: DictType | None = None,
+            query_params: Mapping | None = None,
```
balamurugana, line 167: Do not change `DictType`. It is valid!

minio/minioadmin.py
```diff
@@ -322,6 +323,20 @@ def info(self) -> str:
         )
         return response.data.decode()
 
+    def account_info(self, prefix_usage: bool = False) -> str:
+        """Get usage information for the authenticating account"""
+        query_params = None
+        if prefix_usage:
+            query_params = {
+                "prefix-usage": "true",
+            }
```
balamurugana, line 332:
```suggestion
```

minio/minioadmin.py
```diff
@@ -322,6 +323,20 @@ def info(self) -> str:
...
+        response = self._url_open(
+            "GET",
+            _COMMAND.ACCOUNT_INFO,
+            query_params=query_params,
```
balamurugana, line 336:
```suggestion
            query_params={"prefix_usage": "true"} if prefix_usage else None,
```

balamurugana (changes requested):

minio/minioadmin.py
mhkarimi1383: My IDE says that using mapping is correct :)

Since the type of dict is dynamic

balamurugana: https://github.com/minio/minio-py/blob/master/minio/helpers.py#L82 is what we want.

mhkarimi1383: Fixed

harshavardhana (approved):

PR #1464, opened 2024-12-06
Fix IMDSv2 URL when fetching credentials

setu4993:
I missed 1 change in the final version of https://github.com/minio/minio-py/pull/1437 where the URL to fetch the endpoint must include `"/latest/meta-data/iam/security-credentials/"` before the role name. This adds that back.

Closes https://github.com/minio/minio-py/issues/1459.

### DIFF
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -31,7 +31,7 @@ from abc import ABCMeta, abstractmethod
 from datetime import timedelta
 from pathlib import Path
 from typing import Callable, cast
-from urllib.parse import urlencode, urlsplit, urlunsplit
+from urllib.parse import urlencode, urlsplit
 from xml.etree import ElementTree as ET
 
 import certifi
@@ -44,7 +44,7 @@ except ImportError:
 
 from urllib3.util import Retry, parse_url
 
-from minio.helpers import sha256_hash, url_replace
+from minio.helpers import sha256_hash
 from minio.signer import sign_v4_sts
 from minio.time import from_iso8601utc, to_amz_date, utcnow
 from minio.xml import find, findtext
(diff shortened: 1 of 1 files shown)


minio/credentials/providers.py
```diff
@@ -504,22 +504,18 @@ def retrieve(self) -> Credentials:
...
             res = _urlopen(
                 self._http_client,
                 "GET",
-                urlunsplit(
-                    url_replace(
-                        urlsplit(url),
-                        path="/latest/meta-data/iam/security-credentials/",
-                    ),
-                ),
-                headers=headers,
```
setu4993, line 517: This was unnecessary so I simplified it with L507.

minio/credentials/providers.py
```diff
@@ -504,22 +504,18 @@ def retrieve(self) -> Credentials:
...
+                url,
+                headers=headers
             )
             role_names = res.data.decode("utf-8").split("\n")
             if not role_names:
                 raise ValueError(f"no IAM roles attached to EC2 service {url}")
-            url += "/" + role_names[0].strip("\r")
+            url += role_names[0].strip("\r")
```
setu4993, line 518: The role gets added at the end of the URL that was updated on L507.

setu4993:
@balamurugana : Fixes IMDSv2 that didn't work correctly. Only found out today.

minio/credentials/providers.py
setu4993: This is also in the Go implementation here: https://github.com/minio/minio-go/blob/e337e77c06bbf8a65247fb018e59c1263e4d86e1/pkg/credentials/iam_aws.go#L400-L402

We just missed it in the revisions in https://github.com/minio/minio-py/pull/1437/.

balamurugana:
I am closing on behalf of https://github.com/minio/minio-py/pull/1465

PR #1475, opened 2025-01-09
Fix ExcludedPrefixes in set/get bucket versioning config

vadmeste:
Fix parsing of ExlucdedPrefix in set/get bucket versioning API.
ExcludedPrefixes includes Prefix tag and not ExcludedPrefix as stated in
the S3 spec.

### DIFF
--- a/minio/versioningconfig.py
+++ b/minio/versioningconfig.py
@@ -78,7 +78,7 @@ class VersioningConfig:
             prefix.text
             for prefix in findall(
                 element,
-                "ExcludedPrefixes/ExcludedPrefix",
+                "ExcludedPrefixes/Prefix",
             )
         ]
         exclude_folders = findtext(element, "ExcludeFolders") == "true"
@@ -99,7 +99,7 @@ class VersioningConfig:
         for prefix in self._excluded_prefixes or []:
             SubElement(
                 SubElement(element, "ExcludedPrefixes"),
-                "ExcludedPrefix",
+                "Prefix",
                 prefix,
             )
         if self._exclude_folders:
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -49,7 +49,7 @@ from minio.select import (CSVInputSerialization, CSVOutputSerialization,
                           SelectRequest)
 from minio.sse import SseCustomerKey
 from minio.time import to_http_header
-from minio.versioningconfig import VersioningConfig
+from minio.versioningconfig import SUSPENDED, VersioningConfig
 
 _CLIENT = None  # initialized in main().
 _TEST_FILE = None  # initialized in main().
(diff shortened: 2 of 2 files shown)


harshavardhana (dismissed):

harshavardhana:
@vadmeste lets add a test

minio/versioningconfig.py
```diff
@@ -75,10 +75,10 @@ def fromxml(cls: Type[A], element: ET.Element) -> A:
         status = findtext(element, "Status")
         mfa_delete = findtext(element, "MFADelete")
         excluded_prefixes = [
-            prefix.text
+            findtext(prefix, "Prefix")
             for prefix in findall(
                 element,
-                "ExcludedPrefixes/ExcludedPrefix",
+                "ExcludedPrefixes",
```
balamurugana, line 81: You could do
```
excluded_prefixes = [prefix.text for prefix in findall(element, "ExcludedPrefixes/Prefix"])
```

vadmeste: > You could do

excluded_prefixes = [prefix.text for prefix in findall(element, "ExcludedPrefixes/Prefix"])

this is not working for me at all, probably related to python3 version or something

vadmeste:
```
import xml.etree.ElementTree as ET

def _get_namespace(element: ET.Element) -> str:
    """Exact namespace if found."""
    start = element.tag.find("{")
    if start < 0:
        return ""
    start += 1
    end = element.tag.find("}")
    if end < 0:
        return ""
    return element.tag[start:end]

def findall(element: ET.Element, name: str) -> list[ET.Element]:
    """Namespace aware ElementTree.Element.findall()."""
    namespace = _get_namespace(element)
    return element.findall(
        "ns:" + name if namespace else name,
        {"ns": namespace} if namespace else {},
    )

root = ET.fromstring('<VersioningConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Status>Enabled</Status><ExcludedPrefixes><Prefix>app1/*/_temporary/</Prefix></ExcludedPrefixes><ExcludedPrefixes><Prefix>app2/*/_staging/</Prefix></ExcludedPrefixes><ExcludeFolders>true</ExcludeFolders></VersioningConfiguration>')

print(findall(root, "ExcludedPrefixes/Prefix"))

```

xpath with a namespaced xml does not work well currently

balamurugana: @vadmeste That should be a serious problem. The code is working fine with namespaced XML for other calls. Let me check that.

balamurugana: @vadmeste I sent a fix https://github.com/minio/minio-py/pull/1476. It solves the issue.

balamurugana (approved):

harshavardhana (approved):

PR #1487, opened 2025-03-12
MinioAdmin: add targetuser in add_service_account

maxim-volshin:


minio/minioadmin.py
```diff
@@ -753,6 +753,7 @@ def add_service_account(self,
                             secret_key: str | None = None,
                             name: str | None = None,
                             description: str | None = None,
+                            targetuser: str | None = None,
```
balamurugana, line 756: Move the argument to end i.e. after `status`

balamurugana (changes requested):

PR #1493, opened 2025-04-22
put_object: add append object support

balamurugana:


minio/api.py
```diff
@@ -1890,13 +1893,21 @@ def put_object(
             raise ValueError("retention must be Retention type")
         if not callable(getattr(data, "read")):
             raise ValueError("input data must have callable read()")
+        if write_offset is not None:
+            if write_offset < 0:
+                raise ValueError("write offset should not be negative")
+            if length < 0:
+                raise ValueError("length must be provided for write offset")
+            part_size = length if length > MIN_PART_SIZE else MIN_PART_SIZE
```
kannappanr, line 1901: should this be 5 MiB minimum?

balamurugana: If object size is < 5 MiB

harshavardhana (approved):

PR #1505, opened 2025-08-07
Add region, extra_headers and extra_query_params

balamurugana:
* Now all APIs accepts region, extra_headers and extra_query_params additional arguments for more control.
* Add checksum support


csumisha (changes requested):
The updated method definitions have not been reflected in the sample provided in api.py, documentation, and examples. It has to be fixed.

minio/api.py
```diff
@@ -422,35 +489,44 @@ def _url_open(
...
             no_body_trace: bool = False,
+            region: Optional[str] = None,
+            extra_headers: Optional[HTTPHeaderDict] = None,
+            extra_query_params: Optional[HTTPQueryDict] = None,
     ) -> BaseHTTPResponse:
         """Execute HTTP request."""
-        region = self._get_region(bucket_name)
+        region = self._get_region(
+            bucket_name=bucket_name,
+            region=region,
```
sivanantham-gnanavel, line 514: Missing: extra_headers and extra_query_params

sivanantham-gnanavel, line 514: Missing: extra_headers and extra_query_params

examples/fput_object.py
```diff
@@ -23,74 +23,80 @@
...
+    f"version-id: {result.version_id}",
 )
 
 # Upload data with metadata.
 result = client.fput_object(
-    "my-bucket", "my-object", "my-filename",
+    bucket_name="my-bucket",
+    object_name="my-object",
+    file_path="my-filename",
     metadata={"My-Project": "one"},
```
sivanantham-gnanavel, line 59:
```suggestion
    user_metadata={"My-Project": "one"},
```

sivanantham-gnanavel (changes requested):
Make sure to update API.md and README.md for: 

- Minio() → keyword-only arguments 
- metadata → user_metadata

minio/api.py
```diff
@@ -422,35 +532,44 @@ def _url_open(
...
```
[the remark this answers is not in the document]
balamurugana: The private `_get_region()` method works like below
1. If region was passed via constructor and no region argument is passed, the region passed via constuctor is returned
2. If region was not passed via constructor and region argument is passed, the region passed by argument is returned
3. If region was passed via constructor, region argument is passed and both are not same, `ValueError` is raised
4. If region was passed via constructor, region argument is passed and both are same, the region passed by argument is returned
5. If bucket name is `None` or empty or no credential provider is returned, `us-east-1` is returned
6. If previously cached region value is found for bucket name, cached region is returned
7. If above doesn't find a region, `GetBucketLocation` S3 API is called to fetched/cached/returned the region

We have one place to control region value and `region` argument is passed to `_execute()` is passed to `_get_region()` for the same reason

@sivanantham-gnanavel 
As `GetBucketLocation` S3 API at `_get_region()` is a controlled call, we don't need `extra_headers` and `extra_query_params`

minio/api.py
```diff
@@ -570,71 +718,122 @@ def disable_virtual_style_endpoint(self):
...
+        Create a bucket with region and optional object lock.
+
+        Args:
+            bucket_name (str):
+                Name of the bucket.
+
+            location (Optional[str], default=None):
+                Region in which the bucket is to be created.
+
+            object_lock (bool, default=False):
```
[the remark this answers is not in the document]
balamurugana: No. That will be like `None`, `False`, `True` which doesn't help here. The default value is `False` if we don't pass the argument. This is good enough for `bool` optional argument.

rraulinio (changes requested):
Overall, good job! I do have a few concerns here. Goes as follows:

**(1) Keyword-only arguments**
Every method in the `Minio` class now forces keyword arguments => All user code needs migration => Requires major version bump.

**(2) Changes of `raise ValueError` to `raise TypeError`**
This means that any code implementation catching `ValueError` won't be able to do this anymore.

**(3) Renames of `metadata` to `user_metadata`**
So code using explicitly `metadata` will now (silently) fail.

**(4) No deprecation warnings.**

ebozduman (dismissed):
LGTM

rraulinio (approved):

sivanantham-gnanavel (dismissed):
LGTM

hfudev:
@rraulinio while you're fixing the bug, could you help yank the 7.2.19 version first?

rraulinio:
Feedback here - https://github.com/minio/minio-py/issues/1536#issuecomment-3573899054.

Issue #1496, opened 2025-05-06
Exception raised when using an S3 Access Point


NeutrinoXY:
When using an S3 Access Point, the following error is raised:

```
File "/usr/local/lib/python3.11/site-packages/minio/api.py", line 135, in __init__
    self._base_url = BaseURL(
                     ^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/minio/helpers.py", line 473, in __init__
    self._aws_info, region_in_host = _get_aws_info(
                                     ^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/minio/helpers.py", line 391, in _get_aws_info
    raise ValueError(f"invalid Amazon AWS host {host}")
ValueError: invalid Amazon AWS host abcd-123456789012.s3-accesspoint.us-east-1.amazonaws.com
```

The issue seems to come from a wrong regex string: https://github.com/minio/minio-py/blob/82c6c1af876983b6658ed0c91f3edc7d08d47c7b/minio/helpers.py#L68C1-L68C24

balamurugana:
@NeutrinoXY Yes. `abcd-123456789012.s3-accesspoint.us-east-1.amazonaws.com` is not supported. Feel free to send a PR by fixing the regexp.

balamurugana:
Fixed by https://github.com/minio/minio-py/pull/1505

PR #1508, opened 2025-09-07
[#1496] Dynamic length S3 Access Point Support

ShubhamKaudewar:
- PR Address issue noticed in #1496 
- Update _AWS_S3_PREFIX to allow <accesspoint-name>-<account-id> (where <accesspoint-name> can be longer and include hyphens)
- Support for S3 URL `<accesspoint-name>-<account-id>.s3-accesspoint.<region>.amazonaws.com`


balamurugana:
Taken care in https://github.com/minio/minio-py/pull/1505

PR #1509, opened 2025-09-07
[FEATURE] Batch Bucket Management: Sequential and Parallel Operations

ShubhamKaudewar:
**Hello Community,**
This PR aims to introduce bulk operations for bucket-related tasks. Currently, the MinIO Python SDK supports `make_bucket` and `remove_bucket` functionalities, which allow users to create a single bucket with a specified region and object_lock. This PR proposes extending support to include the following multi-operation functionalities:

- `make_buckets`
- `make_buckets_parallel`
- `remove_buckets`
- `remove_buckets_parallel`

#### Why is this needed?
Users no longer need to write wrapper functions using iterators or loops to create or remove multiple buckets. Parallel support enables bucket creation in isolated threads, improving efficiency and reducing the risk of a single point of failure that could block subsequent bucket creations.

#### Conclusion
Feel free to suggest changes or share feedback. Let me know if this is unnecessary at this time or if it poses any risks or requires additional sanity checks.


balamurugana:
This kind of application logic should be outside of `minio-py`

PR #1522, opened 2025-10-27
fix: composesource object_size property should not call itself

ghostrider77:
Combining data using server-side copy using `client.compose_object("bucket_name", "object_name", sources)` does not work. The reason is that the `compose_object` method iteratively calls the `object_size` property of each `ComposeSource` object and each of them recursively calls itself, resulting in a `RecursionError`.


balamurugana:
A better fix is https://github.com/minio/minio-py/pull/1524

PR #1528, opened 2025-11-13
support optional parameters in LdapIdentityProvider

jkandasa:
Added support for three optional parameters to `LdapIdentityProvider` class

* `duration_seconds`: Specify the number of seconds after which the temporary credentials expire.
* `policy`: Specify the URL-encoded JSON-formatted policy to use as an inline session policy.
* `token_revoke_type`: A string to assign to the generated token that allows the token's access to be revoked prior to expiration.

Document: https://docs.min.io/enterprise/aistor-object-store/developers/security-token-service/assumerolewithldapidentity/

### DIFF
--- a/minio/credentials/providers.py
+++ b/minio/credentials/providers.py
@@ -556,19 +556,29 @@ class LdapIdentityProvider(Provider):
 
     def __init__(
             self,
+            *,
             sts_endpoint: str,
             ldap_username: str,
             ldap_password: str,
+            duration_seconds: Optional[int] = None,
+            policy: Optional[str] = None,
+            token_revoke_type: Optional[str] = None,
             http_client: Optional[PoolManager] = None,
     ):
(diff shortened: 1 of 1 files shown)


minio/credentials/providers.py
```diff
@@ -560,15 +560,25 @@ def __init__(
             ldap_username: str,
             ldap_password: str,
             http_client: Optional[PoolManager] = None,
+            *,
```
balamurugana: Move `*`, after `self`

balamurugana: The `master`  broke many things already. The next release must be `v8.0.0`

jkandasa: @ramondeklein I discussed with @balamurugana, he mentioned that, the current `master` is already having breaking changes.
So next release should be a major release: `v8.0.0`

ramondeklein (approved):

PR #1529, opened 2025-11-13
Incorrect existing checksum headers to both CreateMultipartUpload and UploadPart requests

rraulinio:
Relates to: https://github.com/minio/mint/pull/399.

Remove checksum headers from `CreateMultipartUpload`.

### Problem
Multipart uploads fail with `InvalidPart` error:
> One or more of the specified parts could not be found. The part may not have been uploaded, or the specified entity tag may not match the part's entity tag.

### Root Cause
Incorrectly existing checksum headers (recently added) to both `CreateMultipartUpload` and `UploadPart` requests. This causes ETag mismatch because:

**Why it fails:** When checksum headers are present on `CreateMultipartUpload`, the S3 server establishes expectations for how the final object's ETag should be calculated. When the same checksum headers are then sent with each `UploadPart`, the server calculates part ETags differently than expected from the initial multipart setup. During `CompleteMultipartUpload`, the server validates that uploaded part ETags match what was declared, but they don't match due to the conflicting checksum information, resulting in `InvalidPart` error.

**Per S3 spec:** Checksum headers should only be on individual `UploadPart` requests (where actual data is uploaded), not on `CreateMultipartUpload` (which only initializes the upload with metadata like storage class and encryption).

### Solution
Remove `headers.extend(checksum_headers)` before `CreateMultipartUpload` call. Checksum headers should only be sent with `UploadPart` requests.

**LE: Solution extended to filtering the headers in `CreateMultipartUpload`. On top of that, adding the missing checksum fields in `CompleteMultipartUpload`.**

### DIFF
--- a/minio/api.py
+++ b/minio/api.py
@@ -3298,6 +3298,9 @@ class Minio:
                 object_name=object_name,
                 upload_id=cast(str, upload_id),
                 parts=parts,
+                extra_headers=HTTPHeaderDict(
+                    sse.headers() if isinstance(sse, SseCustomerKey) else None
+                ),
             )
             return ObjectWriteResult.new(
                 headers=upload_result.headers,


rraulinio:
@balamurugana Even if you remove this line from the test - `checksum=Algorithm.CRC32C` - in the `put_object` request, the bug still happens, just realised now.
That's just because `Algorithm.CRC32C` gets picked as default if nothing is requested explicitly.

balamurugana:
@rraulinio We need to fix the bug not to remove checksum support.

rraulinio:
@balamurugana Correct. So we either have this approach that I've just commited, in which we only allow the algo header, but we don't send wrongfully the hash of the first part anymore. Or the approach you suggested, they both solve the problem (I did not test your approach, but I take it will work), which one would you prefer the most and why?

harshavardhana (approved):

PR #1535, opened 2025-11-24
Fix for v7.2.19 related issues

rraulinio:
### Problem
  Commit 6daf366 introduced keyword-only argument separators (`*,`) at the start of function signatures, breaking backward compatibility for all existing client code using positional arguments.

  ### Solution
  - **Removed 78 keyword-only separators** from existing parameters to restore
  backward compatibility
  - **Kept 14 strategic separators** before NEW parameters only (`region`,
  `extra_headers`, `extra_query_params`) to enforce keyword-only usage for new
  additions
  - **Maintained dual parameter support** for the metadata split: kept both old
  `metadata` parameter AND new `headers`/`user_metadata` parameters with
  fallback logic for backward compatibility

  ### Changes
  - **minio/api.py**: Fixed duplicate code, type annotations and added metadata
   parameter to `compose_object`
  - **docs/API.md**: Updated all function signatures to match implementation
  with correct parameter ordering and keyword-only separators
  - **pylintrc**: Added `max-positional-arguments=16` to accommodate
  backward-compatible signatures

  ### Backward Compatibility
  ✅ Old code continues to work: `client.put_object(bucket, obj, data, length)`
  ✅ New parameters require keywords: `client.put_object(...,
  region="us-east-1")`
  ✅ Legacy metadata parameter supported: `metadata={"key": "val"}` automatically falls back to `user_metadata`

### DIFF
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -2153,13 +2153,17 @@ def test_real_world_usage_patterns(log_entry):
             )
 
         # Pattern 3: Copy with minimal args (BROKEN in master)
-        _client.copy_object(bucket_name, "test-copy", CopySource(bucket_name, "test"))
+        _client.copy_object(
+            bucket_name, "test-copy", CopySource(bucket_name, "test"),
+        )
 
-        # Pattern 4: List with prefix positionally (BROKEN in master)
+        # Pattern 4: List with prefix positionally (broken in master)
         list(_client.list_objects(bucket_name, "test"))
 
         # Pattern 5: Mixed positional and keyword (should work)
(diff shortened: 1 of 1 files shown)


harshavardhana:
> I do not prefer this kind of fix and I recommend full backward compatibility. I am not sure how much this PR does. My strong recommendations are

We broke backward compatibility, so this PR reverts to restore the expected functionality.  Not sure what you are saying. 

Once this PR is merged, we will release v7.2.20, so users who were on v7.2.18 won't see anything break when they upgrade.

rraulinio:
Trying a v2 PR as well, yes, then we can pick.

PR #1559, opened 2026-04-03
fix: skip functional tests on HTTP 405/501 responses

harshavardhana:
Use HTTP status codes directly instead of checking error
code strings, so tests are properly skipped when S3 Select
or other APIs return 405 (Method Not Allowed) or
501 (Not Implemented).

### DIFF
--- a/tests/functional/tests.py
+++ b/tests/functional/tests.py
@@ -133,7 +133,7 @@ def _call_test(func, *args, **kwargs):
     try:
         func(log_entry, *args, **kwargs)
     except S3Error as exc:
-        if exc.code == "NotImplemented":
+        if exc.response.status in (405, 501):
             log_entry["alert"] = "Not Implemented"
             log_entry["status"] = "NA"
         else:


tests/functional/tests.py
```diff
@@ -133,7 +133,7 @@ def _call_test(func, *args, **kwargs):
     try:
         func(log_entry, *args, **kwargs)
     except S3Error as exc:
-        if exc.code == "NotImplemented":
+        if exc.response.status in (405, 501):
```
balamurugana, line 136: `Minio` class properly returns `"MethodNotAllowed"` S3 error code for `405` and `501` HTTP status code. We could use it along with `NotImplemented` s3 error code. 

The below handles `NotImplemented` S3 error code irrespective of HTTP status code.
```suggestion
        if exc.code in ["NotImplemented",  "MethodNotAllowed"]:
```

harshavardhana: No, it was implemented on purpose to only look for the exact HTTP status instead of relying on the Code, as the Code can change, and it is not really part of the spec to be the same across all services.

balamurugana: Don't we want to handle `NotImplemented` and `MethodNotAllowed` S3 code for other HTTP status code? If we don't, the test will fail for them

harshavardhana: S3 codes can change anytime, we cannot rely on them. We can rely on http status codes that's why I changed it.

harshavardhana:
PTAL @balamurugana

harshavardhana:
I am taking this in, the PR has gone too long we need this merged and released to unblock some other activities on miniohq/eos

PR #1563, opened 2026-08-04
rdma: reject buffers over the 4 GiB cuObject registration limit

harshavardhana:
## What

A single cuObject registration (`cuMemObjGetDescriptor`, inside libminiocpp) can pin at most **4 GiB**. `RDMAClient.put`/`get` forwarded the buffer size straight to the C API, so an oversized buffer failed **opaquely** in the C layer (`RDMA put/get: <last error>`).

This adds a fail-fast guard: reject `size > 4 GiB` (right after the existing `size <= 0` check) with a clear `ValueError` telling the caller to split the transfer into parts ≤ 4 GiB (multipart upload / ranged read).

- Add `_CUOBJ_MAX_MEMORY_REG_SIZE` (4 GiB) and guard both `put` and `get`.

The substantive handling lives upstream in libminiocpp (already merged); minio-py is a thin ctypes wrapper, so its correct role is a clear guard rather than a workaround. Mirrors the equivalent guards in minio-cpp, minio-rs, and minio-go.

## Testing

`python -m py_compile` clean; `pylint` clean; no import changes (isort/mypy unaffected); longest added line 81 cols.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent RDMA PUT and GET transfers larger than 4 GiB.
  * Transfers exceeding the limit now fail with a clear error message advising you to split them into smaller parts.
  * Transfers up to and including 4 GiB continue to be supported.

### DIFF
--- a/.github/workflows/ci-rdma.yml
+++ b/.github/workflows/ci-rdma.yml
@@ -39,12 +39,19 @@ jobs:
       uses: actions/checkout@v4
       with:
         repository: microsoft/vcpkg
+        ref: "2026.07.29"
         path: "vcpkg"
+
+    # arm64 Linux has no prebuilt vcpkg CMake, so vcpkg falls back to the system
+    # cmake; the apt version (3.28) is too old for vcpkg's SPDX generation
+    # (string(JSON ... STRING_ENCODE)). Provide a recent cmake on PATH.
+    - name: Install recent CMake and Ninja
+      uses: lukka/get-cmake@fffaaafeea488556c2c12dad60690008bc1caacb # v4.4.2
 
     - name: Install system dependencies
       run: |
         sudo apt-get -qy update
-        sudo apt-get -qy install cmake libibverbs-dev librdmacm-dev libnuma-dev
+        sudo apt-get -qy install libibverbs-dev librdmacm-dev libnuma-dev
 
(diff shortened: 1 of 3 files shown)


harshavardhana:
## CodeRabbit Autofix — Applied (02384ab)

**Applied:**
- 🔵 *Add boundary tests for both transfer methods* — new `RDMAOversizeGuardTest` in `tests/unit/rdma_test.py` exercises `RDMAClient.put`/`get` at the inclusive limit and one byte over, via the existing `_FakeLib`. A raw pointer is used so the audit's buffer-vs-length guard doesn't fire first, isolating the 4 GiB size guard. Asserts oversized → `ValueError` with the native put/get **not** called; exact-limit → native called. All 4 pass (7 total in the file).

**Note:** rebased the branch onto latest `master` (picked up the package-wide audit #1562, which added its own `size > inferred` buffer guard and refactored `_buffer_pointer`). Resolved the overlap by keeping **both** guards — the buffer-overflow check and the 4 GiB registration check — in `put` and `get`.

**Skipped:** the embedded Ruff `TRY003` note — the repo doesn't run Ruff and the pre-existing `ValueError` messages already use the same style.

PR #1564, opened 2026-08-12
Use the region passed to get_presigned_url

codechrl:
`get_presigned_url()` accepts a `region` argument, documented as "Region of the bucket to skip auto probing", and overwrites it on the first line of its body:

```python
region = self._get_region(bucket_name=bucket_name)
```

`presigned_get_object()` and `presigned_put_object()` forward `region` to it, so they are affected the same way. #1505 added the argument to the signature without passing it through.

The caller cannot pin the SigV4 credential scope of the generated URL. A client built without `Minio(region=...)` also sends a `GetBucketLocation` request on the first presign for each bucket, which requires network reachability and `s3:GetBucketLocation` permission at the moment the URL is generated. Presigning requires neither otherwise.

`get_presigned_url()` now passes `region` to `_get_region()`, which returns a caller-supplied region unchanged and validates it against `Minio(region=...)`. `region=None` behaves as before. One case that was silent now raises: `Minio(region='us-east-1')` presigning with `region='us-west-2'` gives `ValueError: region must be us-east-1, but passed us-west-2`, matching every other method that takes a region.

## Summary by CodeRabbit

- **Bug Fixes**
  - Presigned URLs now correctly use an explicitly provided region when generating signed object links.
  - Improved regional signing accuracy, including support for regions such as `us-west-2`.
  - Added validation to detect conflicts between the requested region and the client’s configured region.

- **Tests**
  - Added coverage for explicit-region presigned GET URLs, regional signing scope, region probing, and conflicting region settings.

### DIFF
--- a/minio/minio.py
+++ b/minio/minio.py
@@ -4911,7 +4911,7 @@ class Minio:
         if expires.total_seconds() < 1 or expires.total_seconds() > 604800:
             raise ValueError("expires must be between 1 second to 7 days")
 
-        region = self._get_region(bucket_name=bucket_name)
+        region = self._get_region(bucket_name=bucket_name, region=region)
         query_params = HTTPQueryDict()
         if version_id:
             query_params["versionId"] = version_id


tests/unit/presigned_get_object_test.py
```diff
@@ -63,3 +65,38 @@ def test_can_include_response_headers(self):
         self.assertIn('inline', r)
         self.assertIn('test.pdf', r)
         self.assertIn('application%2Fpdf', r)
+
+    @mock.patch('urllib3.PoolManager')
+    def test_region_is_used(self, mock_connection):
```
balamurugana: We do not need this test.

balamurugana (changes requested):

codechrl:
Dropped the tests.

balamurugana (approved):

PR #1567, opened 2026-08-16
Keep the region on path-style transfer-acceleration URLs

ckarnell:
Transfer acceleration has no path-style form, so `_build_aws_url` strips `-accelerate` and falls back to the plain `s3.` host:

```python
netloc = s3_prefix
if "s3-accelerate" in s3_prefix:
    ...
    if enforce_path_style:
        netloc = netloc.replace("-accelerate", "", 1)
...
if "s3-accelerate" not in s3_prefix:
    netloc += region + "."
```

The strip rewrites `netloc`, but the guard underneath re-tests `s3_prefix`, which still says `s3-accelerate.`. So the region never gets appended.

`enforce_path_style` is set for CreateBucket and GetBucketLocation, and for any bucket name containing a dot over HTTPS. With `Minio("s3-accelerate.amazonaws.com", region="us-west-2")`:

```
make_bucket           -> s3.amazonaws.com          (want s3.us-west-2.amazonaws.com)
GetBucketLocation     -> s3.amazonaws.com          (want s3.us-west-2.amazonaws.com)
get_object            -> my-bucket.s3-accelerate.amazonaws.com   (correct, unchanged)
```

`s3.amazonaws.com` is us-east-1. So those two calls hit the wrong region while the request is signed for another one. The same client against `s3.amazonaws.com` builds `s3.us-west-2.amazonaws.com`, which is what the accelerate fallback should match.

Before #1289 this line read `if enforce_path_style or not self._accelerate_host_flag`. I put that term back, so this restores old behaviour instead of inventing any.

Three tests added to `helpers_test.py`. The two path-style ones fail on master and pass with the change, and the third pins that virtual-style accelerate keeps its host either way. Unit tests go from 121 passed, 2 skipped to 124 passed, 2 skipped. `pylint`, `mypy minio`, `autopep8 --diff` and `isort --diff` are all clean.

I have not run this against a real accelerate endpoint, so the evidence is the host string the builder produces, not a live request.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected accelerated S3 endpoint handling when path-style addressing is enabled.
  * Region-specific endpoints now resolve correctly, including dual-stack configurations.
  * Preserved existing behavior for virtual-hosted S3 Accelerate endpoints.

* **Tests**
  * Added coverage for regional, virtual-hosted, and dual-stack accelerated endpoint scenarios.

### DIFF
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -472,7 +472,7 @@ class BaseURL:
 
         if aws_info["dualstack"]:
             netloc += "dualstack."
-        if "s3-accelerate" not in s3_prefix:
+        if enforce_path_style or "s3-accelerate" not in s3_prefix:
             netloc += region + "."
         netloc += domain_suffix
 
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -468,11 +468,12 @@ class BaseURL:
                     f"for accelerate endpoint"
                 )
             if enforce_path_style:
-                netloc = netloc.replace("-accelerate", "", 1)
+                s3_prefix = netloc.replace("-accelerate", "", 1)
+                netloc = s3_prefix
 
         if aws_info["dualstack"]:
             netloc += "dualstack."
-        if enforce_path_style or "s3-accelerate" not in s3_prefix:
+        if "s3-accelerate" not in s3_prefix:
             netloc += region + "."
         netloc += domain_suffix
 
--- a/tests/unit/helpers_test.py
+++ b/tests/unit/helpers_test.py
@@ -1895,3 +1895,45 @@ class BaseURLTests(TestCase):
                 ),
             )
             self.assertEqual(str(url), case.result)
+
+    def test_accelerate_path_style_keeps_region(self):
+        # Path style is forced for CreateBucket, GetBucketLocation and dotted
+        # bucket names. Accelerate has no path-style form, so the host falls
+        # back to "s3." and that form is region specific.
+        base_url = BaseURL("https://s3-accelerate.amazonaws.com", "us-west-2")
+
+        self.assertEqual(
+            base_url.build(
+                method="PUT", region="us-west-2", bucket_name="my-bucket",
+            ).netloc,
+            "s3.us-west-2.amazonaws.com",
+        )
+        self.assertEqual(
+            base_url.build(
+                method="GET", region="us-west-2", bucket_name="my-bucket",
+                query_params={"location": ""},
+            ).netloc,
+            "s3.us-west-2.amazonaws.com",
+        )
+
+    def test_accelerate_virtual_style_is_unchanged(self):
+        base_url = BaseURL("https://s3-accelerate.amazonaws.com", "us-west-2")
+
+        self.assertEqual(
+            base_url.build(
+                method="GET", region="us-west-2", bucket_name="my-bucket",
+                object_name="path/to/my/object",
+            ).netloc,
+            "my-bucket.s3-accelerate.amazonaws.com",
+        )
+
+    def test_accelerate_path_style_keeps_region_with_dualstack(self):
+        base_url = BaseURL("https://s3-accelerate.amazonaws.com", "us-west-2")
+        base_url.dualstack_host_flag = True
+
+        self.assertEqual(
+            base_url.build(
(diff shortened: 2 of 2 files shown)


minio/helpers.py
```diff
@@ -472,7 +472,7 @@ def _build_aws_url(
 
         if aws_info["dualstack"]:
             netloc += "dualstack."
-        if "s3-accelerate" not in s3_prefix:
+        if enforce_path_style or "s3-accelerate" not in s3_prefix:
```
balamurugana: The fix is logically incorrect and functionally working ATM. The actual fix is
```diff
diff --git a/minio/helpers.py b/minio/helpers.py
index a661fe9..83f5537 100644
--- a/minio/helpers.py
+++ b/minio/helpers.py
@@ -468,7 +468,8 @@ class BaseURL:
                     f"for accelerate endpoint"
                 )
             if enforce_path_style:
-                netloc = netloc.replace("-accelerate", "", 1)
+                s3_prefix = netloc.replace("-accelerate", "", 1)
+                netloc = s3_prefix
 
         if aws_info["dualstack"]:
             netloc += "dualstack."
```

tests/unit/helpers_test.py
```diff
@@ -1895,3 +1895,45 @@ def test_aws_object_build(self):
                 ),
             )
             self.assertEqual(str(url), case.result)
+
```
balamurugana: Add unit test for `s3-accelarate` and `enforce_path_style` combo like `test_aws_bucket_build(self)` test

balamurugana (changes requested):

ckarnell:
Mine left `s3_prefix` saying `s3-accelerate.` while `netloc` had already become `s3.`, so the check below was reading a variable that no longer described the host. Adding `enforce_path_style` to that condition papers over the disagreement instead of removing it. Yours keeps the two in step and the existing check then works with no special case, which is the better fix.

Pushed your diff as-is.

Also added the accelerate plus `enforce_path_style` case to the build tests, in the shape of `test_aws_bucket_build`:

```
("https://s3-accelerate.amazonaws.com", None)              -> https://s3.us-east-1.amazonaws.com/my-bucket
("https://s3-accelerate.amazonaws.com", "ap-south-1a")     -> https://s3.ap-south-1a.amazonaws.com/my-bucket
("https://s3-accelerate.dualstack.amazonaws.com", "us-west-2") -> https://s3.dualstack.us-west-2.amazonaws.com/my-bucket
```

On master the first of those gives `https://s3.amazonaws.com/my-bucket`, so the case fails on the missing region, not on formatting. Unit tests are 122 passed, 2 skipped. `pylint`, `mypy`, `autopep8 --diff` and `isort --diff` are clean.

The virtual-style accelerate cases already in `test_aws_bucket_build` still pass untouched, which is the half I wanted to be sure of.

minio/helpers.py
```diff
@@ -468,11 +468,12 @@ def _build_aws_url(
...
                 )
             if enforce_path_style:
-                netloc = netloc.replace("-accelerate", "", 1)
+                s3_prefix = netloc.replace("-accelerate", "", 1)
+                netloc = s3_prefix
 
         if aws_info["dualstack"]:
             netloc += "dualstack."
-        if "s3-accelerate" not in s3_prefix:
+        if enforce_path_style or "s3-accelerate" not in s3_prefix:
```
balamurugana: Why do we need this?

tests/unit/helpers_test.py
```diff
@@ -1895,3 +1925,45 @@ def test_aws_object_build(self):
...
+    def test_accelerate_path_style_keeps_region(self):
```
balamurugana: Why do we need this?

balamurugana (commented):
AI makes mistakes. It is required to review the code before submitting. Have you tested this PR with AWS S3 accelerate endpoint?

ckarnell:
Once s3_prefix carries the corrected host, the enforce_path_style term in that check is dead weight, so I dropped it and the condition reads plainly. I also collapsed the three accelerate tests into the single build-style one you asked for. Force-pushed, so the branch is now your fix plus that one test.

balamurugana (approved):

